PR #16, opened 2020-03-18
Expose data and params kwargs for request helper

henrylamchan:
# Overview
Expose data and params for the request helper as pass through kwargs. This'd allow to make the oauth post call with query params. Straightforward abstraction that I hope doesn't cause much confusion in the future.

### DIFF
--- a/workos/utils/request.py
+++ b/workos/utils/request.py
@@ -34,7 +34,13 @@ class RequestHelper(object):
         return self.base_api_url.format(path)
 
     def request(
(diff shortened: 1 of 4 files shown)


workos/utils/request.py
```diff
@@ -60,9 +66,9 @@ def request(
 
         request_fn = getattr(requests, method)
         if method == REQUEST_METHOD_GET:
-            response = request_fn(url, headers=headers, params=params)
+            response = request_fn(url, headers=headers, data=data, params=params)
```
maxchehab: Should we be sending any data during a GET request?

PR #20, opened 2020-04-17
Use `list` when retrieving multiple records in Directory Sync

rjvani:


workos/directory_sync.py
```diff
@@ -85,7 +85,7 @@ def get_directory_user(self, directory_endpoint_id, directory_user_id):
             token=workos.api_key,
         )
 
-    def get_directory_user_groups(self, directory_endpoint_id, directory_user_id):
+    def list_directory_user_groups(self, directory_endpoint_id, directory_user_id):
```
marktran, line 88: @rjvani @maxchehab 

How about removing "directory" from all of these method names? Seems redundant since the class is already named `DirectorySync`.

`self.directory_sync.list_directory_users` => `self.directory_sync.list_users`

maxchehab: Would `self.directory_sync.list_directories` change to `self.directory_sync.list`?

marktran: I think that can be `self.directory_sync.list_directories`

maxchehab: Not opposed @rjvani what do you think?

rjvani: That sounds good to me 👍 I'll update in another PR.

PR #23, opened 2020-04-22
Rename endpoints when fetching Directory Groups and Users

rjvani:
* Re-factor `Directory Endpoint` to `Directory`
* Drop `_id` from Directory Sync kwargs
* Uses `/directory_users` instead of `/directories/:id/users`
* Uses `/directory_groups` instead of `/directories/:id/groups`
* Remove retrieving Directory Users Groups

### DIFF
--- a/workos/directory_sync.py
+++ b/workos/directory_sync.py
@@ -19,86 +19,76 @@ class DirectorySync(object):
         return self._request_helper
 
     def list_users(
-        self, directory_endpoint_id, limit=RESPONSE_LIMIT, before=None, after=None
+        self, directory=None, group=None, limit=RESPONSE_LIMIT, before=None, after=None
     ):
-        """Gets a list of provisioned Users for a Directory Endpoint.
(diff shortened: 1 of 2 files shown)


workos/directory_sync.py
```diff
@@ -19,86 +19,76 @@ def request_helper(self):
...
-                directory_endpoint_id=directory_endpoint_id
-            ),
+            "directory_groups",
             method=REQUEST_METHOD_GET,
             params=params,
             token=workos.api_key,
         )
 
-    def get_user(self, directory_endpoint_id, directory_user_id):
+    def get_user(self, directory, directory_user):
```
henrylamchan, line 79: I have no in depth knowledge of the app side changes or dsync in general but... is getting a user not doable via the `directory_users` endpoint?

rjvani: You definitely can, but the `directory_users` endpoint gives back a paginated list. I think it's still good to have a way to retrieve a single record.

PR #33, opened 2020-07-22
Support for GET /events endpoint

henrylamchan:
- Refactor capture_and_mock_requests
- Add Resource base class
- Add WorkOSEvent and WorkOSEventAction Resources
- Add support for GET /events endpoint

### DIFF
--- a/workos/audit_trail.py
+++ b/workos/audit_trail.py
@@ -88,6 +88,32 @@ class AuditTrail(object):
         occurred_at_lte=None,
         search=None,
     ):
+        """Filter for Audit Trail Events.
+
+        Kwargs:
+            before (str) - Event ID to look before
+            after (str) - Event ID to look after
+            limit (int) - Number of Events to return
+            group (str|list) - Group or groups to filter for
+            action (str|list) - Action or actions to filter for
+            action_type (str|list) - Action type or types to filter for
+            actor_name (str|list) - Actor name or name to filter for
+            actor_id (str|list) - Actor ID or IDs to filter for
+            target_name (str|list) - Target name or names to filter for
+            target_id (str|list) - Target ID or IDs to filter for
+            occurred_at (str) - ISO-8601 datetime of when an event occurred
+            occurred_at_gt (str) - ISO-8601 datetime of when an event occurred after
+            occurred_at_gte (str) - ISO-8601 datetime of when an event occurred at or after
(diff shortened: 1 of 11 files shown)


workos/audit_trail.py
```diff
@@ -48,21 +50,126 @@ def create_event(self, event, idempotency_key=None):
...
-        return self.request_helper.request(
+        self.request_helper.request(
             EVENTS_PATH,
             method=REQUEST_METHOD_POST,
             params=event,
             headers=headers,
             token=workos.api_key,
         )
+
+        return True
```
CelestineKao, line 72: why return True instead of the response? looking at the other methods, it looks like the norm is to return the response, ex.`SSO.create_connection`?

workos/audit_trail.py
```diff
@@ -48,21 +50,126 @@ def create_event(self, event, idempotency_key=None):
...
+        else:
+            if occurred_at_gte is not None:
+                params["occurred_at_gte"] = occurred_at_gte
+            elif occurred_at_gt is not None:
+                params["occurred_at_gt"] = occurred_at_gt
+
+            if occurred_at_lte is not None:
+                params["occurred_at_lte"] = occurred_at_lte
+            elif occurred_at_lt is not None:
+                params["occurred_at_lt"] = occurred_at_lt
```
CelestineKao, line 160: is this validation handled by our api? if so, is it necessary to do so here?

workos/resources/base.py
```diff
@@ -0,0 +1,36 @@
+class WorkOSBaseResource(object):
+    """Representation of a WorkOS Resource as returned through the API.
+    
+    Attributes:
+        OBJECT_FIELDS (list): List of fields a Resource is comprised of.
+    """
+
+    OBJECT_FIELDS = []
```
CelestineKao, line 8: hmm, should this be a class variable, as opposed to an instance variable instead?

workos/audit_trail.py
henrylamchan: Don't think there are stories for this yet but we've been wanting to create `Resources` for everything. I also sorta don't like returning the raw response as well which was why I opted for this

workos/audit_trail.py
henrylamchan: The validation isn't on the app side. I also think the database should query for the right thing either way given the extraneous args so this is moreso me wanting a cleaner set of filters passed over the wire

PR #37, opened 2020-09-11
Add Portal module with list_organizations function

willmanduffy:
* Returns a list of the organizations accessible to the provided
  API key. Includes pagination functionality via the list_metadata
  property.

### DIFF
--- a/tests/test_directory_sync.py
+++ b/tests/test_directory_sync.py
@@ -8,7 +8,7 @@ from workos.directory_sync import DirectorySync
 from workos.utils.request import RESPONSE_TYPE_CODE
 
 
-class TestSSO(object):
+class TestDirectorySync(object):
     @pytest.fixture(autouse=True)
     def setup(self, set_api_key):
         self.directory_sync = DirectorySync()
--- a/workos/portal.py
+++ b/workos/portal.py
@@ -1,0 +1,44 @@
+import workos
+from workos.utils.request import RequestHelper, REQUEST_METHOD_GET
+from workos.utils.validation import PORTAL_MODULE, validate_settings
+
+RESPONSE_LIMIT = 10
+
+
+class Portal(object):
+    @validate_settings(PORTAL_MODULE)
+    def __init__(self):
+        pass
+
+    @property
+    def request_helper(self):
+        if not getattr(self, "_request_helper", None):
+            self._request_helper = RequestHelper()
+        return self._request_helper
+
+    def list_organizations(
+        self, domains=None, limit=RESPONSE_LIMIT, before=None, after=None
+    ):
+        """Retrieve a list of organizations that have connections configured within your WorkOS dashboard.
+
+        Kwargs:
+            domains (list): Filter organizations to only return those that are associated with the provided domains. (Optional)
+            limit (int): Maximum number of records to return. (Optional)
+            before (str): Pagination cursor to receive records before a provided Organization ID. (Optional)
+            after (str): Pagination cursor to receive records after a provided Organization ID. (Optional)
+
+        Returns:
+            dict: Organizations response from WorkOS.
+        """
+        params = {
+            "domains": domains,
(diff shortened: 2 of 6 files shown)


tests/test_directory_sync.py
```diff
@@ -8,7 +8,7 @@
 from workos.utils.request import RESPONSE_TYPE_CODE
 
 
-class TestSSO(object):
+class TestDirectorySync(object):
```
willmanduffy, line 11: Small unrelated change. I noticed this typo while I was copying other people's code :)

maxchehab (approved):
lgtm, but I'll defer to @henrylamchan or @rjvani for the 👍

workos/portal.py
```diff
@@ -0,0 +1,44 @@
...
+        if not getattr(self, "_request_helper", None):
+            self._request_helper = RequestHelper()
+        return self._request_helper
+
+    def list_organizations(
+        self, domain=None, limit=RESPONSE_LIMIT, before=None, after=None
+    ):
+        """Retrieve a list of organizations that have connections configured within your WorkOS dashboard.
+
+        Args:
```
henrylamchan, line 24: Kwargs not Args

workos/portal.py
```diff
@@ -0,0 +1,44 @@
...
+        pass
+
+    @property
+    def request_helper(self):
+        if not getattr(self, "_request_helper", None):
+            self._request_helper = RequestHelper()
+        return self._request_helper
+
+    def list_organizations(
+        self, domain=None, limit=RESPONSE_LIMIT, before=None, after=None
```
henrylamchan, line 20: Looks like the params expects on API are domains (plural not singular). Does this work as is?

henrylamchan (changes requested):

workos/portal.py
willmanduffy: This is a great catch @henrylamchan looks like my original spec was wrong and I tracked this bug across multiple SDKs. I'll fix here and give a fix to Node and Ruby before releasing them. Good eye!

workos/portal.py
```diff
@@ -0,0 +1,44 @@
...
+            domains (list): Filter organizations to only return those that are associated with the provided domains. (Optional)
+            limit (int): Maximum number of records to return. (Optional)
+            before (str): Pagination cursor to receive records before a provided Organization ID. (Optional)
+            after (str): Pagination cursor to receive records after a provided Organization ID. (Optional)
+
+        Returns:
+            dict: Organizations response from WorkOS.
+        """
+        params = {
+            "domains": domains,
```
henrylamchan, line 34: I think a gotcha that I ran into when implementing get events is that the way query params are serialized from the python requests library.. is valid but nestjs throws up on validation because a list with one item is deserialized as just a string. that being said, you may want to name the key for `domains` as `domains[]` because of that

willmanduffy: Hmmm weird, I did a full request to the API and it seemed to handle it just fine. Happy to swap this up though!

henrylamchan: Ah, if it works its all good /shrug

willmanduffy: Just double checked. I added a real API and did this diff:

```diff
     def test_list_organizations(self, mock_organizations, mock_request_method):
-        mock_response = Response()
-        mock_response.status_code = 200
-        mock_response.response_dict = mock_organizations
-        mock_request_method("get", mock_response, 200)
-        response = self.portal.list_organizations()
-        assert response.status_code == 200
-        assert len(response.response_dict["data"]) == 2
+        response = self.portal.list_organizations(domains=["example2.com"])
+        assert len(response["data"]) == 1
+        assert response["data"] == True
```

First test passes and second test fails but shows the response I expected. Feeling good it works so I'm going to dismiss the blocking review and merge!

henrylamchan (approved):

PR #39, opened 2020-09-14
Add generate_link function to Portal module

willmanduffy:
* Adds an SDK method to generate an Admin Portal link.

### DIFF
--- a/workos/portal.py
+++ b/workos/portal.py
@@ -3,6 +3,7 @@ from workos.utils.request import RequestHelper, REQUEST_METHOD_GET, REQUEST_METH
 from workos.utils.validation import PORTAL_MODULE, validate_settings
 
 ORGANIZATIONS_PATH = "organizations"
+PORTAL_GENERATE_PATH = "portal/generate_link"
 RESPONSE_LIMIT = 10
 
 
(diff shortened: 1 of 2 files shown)


workos/portal.py
```diff
@@ -16,6 +17,47 @@ def request_helper(self):
...
+            token=workos.api_key,
+        )
+
+    def generate_link(self, intent, organization, return_url=None):
+        """Generate a link to grant access to an organization's Admin Portal
+
+        Args:
+            intent (str): The access scope for the generated Admin Portal link. Valid values are: ["sso"]
+            organization (string): The ID of the organization the Admin Portal link will be generated for
+            return_url (str): The URL that the end user will be redirected to upon exiting the generated Admin Portal. If none is provided, the default redirect link set in your WorkOS Dashboard will be used. (Optional)
```
henrylamchan, line 44: Could you move `return_url` into a `Kwargs` section?

henrylamchan: Probably a discussion for some other time but... it's called return_url but we're using the default reDirect_uri if one is not supplied?

workos/portal.py
```diff
@@ -16,6 +17,47 @@ def request_helper(self):
...
+        Returns:
+            str:  URL to redirect a User to to access an Admin Portal session
+        """
+        params = {
+            "intent": intent,
+            "organization": organization,
+            "return_url": return_url,
+        }
+        return self.request_helper.request(
+            "portal/generate_link",
```
henrylamchan, line 55: Could you create a constant for this for consistency?

workos/portal.py
```diff
@@ -16,6 +17,47 @@ def request_helper(self):
             self._request_helper = RequestHelper()
         return self._request_helper
 
+    def create_organization(self, organization):
```
henrylamchan, line 20: Curious if Organizations are going to be used for anything else in the future. If so, would it make sense to put it into a Portal module or should it go somewhere else?

willmanduffy: We had this discussion as we kicked things off and it was decided that while it's not the best future home it is the one most consistent with the approach we've taken so far. In the future we might re-organize things to a more resource based approach, but that would have enough breaking changes we'd want it to be a major version.

workos/portal.py
willmanduffy: Had the same question as I was writing this. @maxchehab?

maxchehab: We are deciding to keep this as is. [See slack convo](https://work-os.slack.com/archives/C01B0NHQQ6M/p1600113895015200).

willmanduffy:
@henrylamchan I've updated with all the outstanding changes. Ready for a another review!

maxchehab (approved):

PR #41, opened 2020-09-16
Add MagicLink support through passwordless module

zlwaterfield:
Add Passwordless support for MagicLink. Adds two methods:
- `create_session` - generate a passwordless session for a user from their email
- `send_session` - send an email to a user with a Magic Link confirmation URL

### DIFF
--- a/workos/passwordless.py
+++ b/workos/passwordless.py
@@ -1,0 +1,59 @@
+import workos
+from workos.utils.request import RequestHelper, REQUEST_METHOD_POST
+from workos.utils.validation import PASSWORDLESS_MODULE, validate_settings
+
+
+class Passwordless(object):
+    """Offers methods through the WorkOS Passwordless service."""
+
+    @validate_settings(PASSWORDLESS_MODULE)
+    def __init__(self):
+        pass
+
+    @property
+    def request_helper(self):
+        if not getattr(self, "_request_helper", None):
+            self._request_helper = RequestHelper()
+        return self._request_helper
+
+    def create_session(self, session_options):
+        """Create an Passwordless Session.
+
+        Args:
+            session_options (dict) - An session options object
+                session_options[email] (str): The email of the user to authenticate.
+                session_options[state] (str): Optional parameter that the redirect
+                    URI received from WorkOS will contain. The state parameter
+                    can be used to encode arbitrary information to help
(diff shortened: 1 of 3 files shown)


zlwaterfield:
Tested locally using a sandbox API key and it is working.

workos/passwordless.py
```diff
@@ -0,0 +1,59 @@
...
+            "passwordless/sessions",
+            method=REQUEST_METHOD_POST,
+            params=session_options,
+            token=workos.api_key,
+        )
+
+    def send_session(self, session_id):
+        """Send a Passwordless Session via email.
+
+        Note, either 'directory' or 'user' must be provided.
```
rjvani, line 46: Can you remove this?

workos/passwordless.py
```diff
@@ -0,0 +1,59 @@
...
+
+        Args:
+            session_id (str): The unique identifier of the Passwordless
+      				Session to send an email for.
+
+        Returns:
+            dict: {"success": true} if successful
```
rjvani, line 53: In our Audit Trail module, we just return `True` directly. I think we should be consistent. @henrylamchan thoughts?

henrylamchan: Im down for just returning `True`. We're pretty inconsistent now with the python SDK but we want to move in a direction where we're not passing back a raw response.

rjvani: Yeah I agree with that. Let's return `True` here as well.

workos/passwordless.py
zlwaterfield: - [ ] remove unnecessary comment

workos/passwordless.py
zlwaterfield: Ok cool, I can do that.

- [ ] return true instead of the `dict`

workos/passwordless.py
```diff
@@ -0,0 +1,59 @@
...
+        pass
+
+    @property
+    def request_helper(self):
+        if not getattr(self, "_request_helper", None):
+            self._request_helper = RequestHelper()
+        return self._request_helper
+
+    def create_session(self, session_options):
+        """Create an Passwordless Session.
```
rjvani, line 20: s/an/a

rjvani (approved):

PR #43, opened 2020-10-13
Expose `connection_id` on Profile Objects

jthodge:
This PR introduces the following changes:

* Passes the `connection_id` Profile object attribute through to the SDK. We currently expose `connection_id` via the `/sso/token` API endpoint.
* Updates the SDK version to `0.8.0`.

### DIFF
--- a/tests/test_sso.py
+++ b/tests/test_sso.py
@@ -26,6 +26,7 @@ class TestSSO(object):
             "email": "demo@workos-okta.com",
             "first_name": "WorkOS",
             "last_name": "Demo",
+            "connection_id": "conn_01EMH8WAK20T42N2NBMNBCYHAG"
             "connection_type": "OktaSAML",
             "idp_id": "00u1klkowm8EGah2H357",
             "raw_attributes": {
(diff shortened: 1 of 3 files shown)


workos/__about__.py
```diff
@@ -12,7 +12,7 @@
 
 __package_url__ = "https://github.com/workos-inc/workos-python"
 
-__version__ = "0.7.0"
+__version__ = "0.8.0"
```
rjvani, line 15: Can we make this 0.7.1?

rjvani (approved):

workos/__about__.py
jthodge: @rjvani I was thinking it's a minor bump since [it's a backwards-compatible feature addition instead of a fix](https://github.com/workos-inc/workos/wiki/SDK-Release-Process#versioning-policy).

tests/test_sso.py
```diff
@@ -26,6 +26,7 @@ def mock_profile(self):
             "email": "demo@workos-okta.com",
             "first_name": "WorkOS",
             "last_name": "Demo",
+            "connection_id": "conn_01EMH8WAK20T42N2NBMNBCYHAG"
```
henrylamchan, line 29: Looks like you need a comma here

henrylamchan (changes requested):

henrylamchan (approved):

PR #50, opened 2021-01-07
Feature/ch4527-32/sdk-updates-for-dashboard2

loferris:
This is a feature branch addressing changes in the SDK for Audit Trail, Directory Sync, and SSO.

This branch contains the following tickets:
ch4527
ch4528
ch4529
ch4530
ch4531
ch4532

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -23,6 +23,13 @@ def set_api_key(monkeypatch):
 def set_project_id(monkeypatch):
     monkeypatch.setattr(workos, "project_id", "project_b27needthisforssotemxo")
 
+@pytest.fixture
+def set_client_id(monkeypatch):
+    monkeypatch.setattr(workos, "client_id", "client_b27needthisforssotemxo")
+
+@pytest.fixture
+def set_api_key_and_client_id(set_api_key, set_client_id):
+    pass
 
 @pytest.fixture
 def set_api_key_and_project_id(set_api_key, set_project_id):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,18 +21,26 @@ def set_api_key(monkeypatch):
 
 @pytest.fixture
 def set_environment_id(monkeypatch):
-    monkeypatch.setattr(workos, "project_id", "environment_b27needthisforssotemxo")
+    monkeypatch.setattr(workos, "environment_id", "environment_b27needthisforssotemxo")
 
 @pytest.fixture
 def set_client_id(monkeypatch):
     monkeypatch.setattr(workos, "client_id", "client_b27needthisforssotemxo")
 
 @pytest.fixture
+def set_project_id(monkeypatch):
+    monkeypatch.setattr(workos, "project_id", "project_b27needthisforssotemxo")
+
+@pytest.fixture
 def set_api_key_and_client_id(set_api_key, set_client_id):
     pass
 
 @pytest.fixture
-def set_api_key_and_project_id(set_api_key, set_environment_id):
+def set_api_key_and_project_id(set_api_key, set_project_id):
+    pass
+
+@pytest.fixture
+def set_api_key_and_environment_id(set_api_key, set_environment_id):
     pass
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -20,11 +20,6 @@ def set_api_key(monkeypatch):
 
 
 @pytest.fixture
-def set_environment_id(monkeypatch):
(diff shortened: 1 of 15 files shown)


loferris:
(I'm trying to figure out the formatting issue - I think I'm on the wrong version of pip or black for semaphore)

loferris:
> (I'm trying to figure out the formatting issue - I think I'm on the wrong version of pip or black for semaphore)

Sorted! Got semaphore's version of formatter from the raw log.

tests/conftest.py
```diff
@@ -19,16 +19,36 @@ def set_api_key(monkeypatch):
     monkeypatch.setattr(workos, "api_key", "sk_abdsomecharactersm284")
 
 
+@pytest.fixture
+def set_environment_id(monkeypatch):
```
rjvani, line 23: Do we need this?

tests/conftest.py
```diff
@@ -19,16 +19,36 @@ def set_api_key(monkeypatch):
...
+    pass
+
+
 @pytest.fixture
 def set_api_key_and_project_id(set_api_key, set_project_id):
     pass
 
 
+@pytest.fixture
+def set_api_key_and_environment_id(set_api_key, set_environment_id):
```
rjvani, line 48: Do we need this?

tests/test_audit_trail.py
```diff
@@ -10,7 +10,7 @@
 
 class TestAuditTrail(object):
     @pytest.fixture(autouse=True)
-    def setup(self, set_api_key_and_project_id):
+    def setup(self, set_api_key_and_environment_id):
```
rjvani, line 13: I think we only need to set API Key here.

tests/test_client.py
```diff
@@ -13,10 +13,10 @@ def setup(self):
         client._portal = None
         client._sso = None
 
-    def test_initialize_sso(self, set_api_key_and_project_id):
+    def test_initialize_sso(self, set_api_key_and_client_id):
         assert bool(client.sso)
 
-    def test_initialize_audit_log(self, set_api_key_and_project_id):
+    def test_initialize_audit_log(self, set_api_key_and_environment_id):
```
rjvani, line 19: I think we only need to set API Key here.

workos/__init__.py
```diff
@@ -5,4 +5,6 @@
 
 api_key = os.getenv("WORKOS_API_KEY")
 project_id = os.getenv("WORKOS_PROJECT_ID")
+client_id = os.getenv("WORKOS_CLIENT_ID")
+environment_id = os.getenv("WORKOS_ENVIRONMENT_ID")
```
rjvani, line 9: We can remove `environment_id` here.

workos/sso.py
```diff
@@ -50,7 +50,8 @@ def get_authorization_url(
             str: URL to redirect a User to to begin the OAuth workflow with WorkOS
         """
         params = {
-            "client_id": workos.project_id,
+            "client_id": workos.client_id,
+            "project_id": workos.project_id,
```
rjvani, line 54: Can we raise a warning if `project_id` is used here? It can be similar to the call made [here](https://github.com/workos-inc/workos-python/blob/master/workos/sso.py#L118-L121).

rjvani (changes requested):

tests/conftest.py
loferris: With the clarification of not needing env_id for the AuditTrail test, no we wouldn't need this fixture!

loferris: Actually, @rjvani, if this is true for passwordless as well (https://github.com/workos-inc/workos-python/blob/feature/ch4527-32/sdk-updates-for-dashboard2/tests/test_passwordless.py), we could remove project_id fixtures as well!

workos/sso.py
rjvani: We don't need to pass `project_id` in the parameters here (an example call can be seen in the docs [here](https://workos.com/docs/reference/sso/redirect-uri/get)). `workos.project_id` or `workos.client_id` should be assigned to `params['client_id']`, but if `workos.project_id` is non-null, we should raise the warning.

workos/sso.py
```diff
@@ -90,12 +97,19 @@ def get_profile(self, code):
             WorkOSProfile: WorkOSProfile object representing the User
         """
         params = {
-            "client_id": workos.project_id,
+            "client_id": workos.client_id,
+            "project_id": workos.project_id,
```
rjvani, line 101: Same comment as above.

rjvani (changes requested):

rjvani (approved):
Awesome! 💯

PR #57, opened 2021-02-23
feat(workos-python): Implement `__str__` for `BaseRequestException`

rohanjadvani:
Implement `__str__` for a human-readable format as exceptions come up.


jthodge (approved):
Nice! This'll be a much better improvement to the current stack trace we surface `:chef-kiss:`

PR #72, opened 2021-05-14
Add `get_organization` method

erin-workos:


erin-workos:
@workos-inc/python I will be out on Friday 5/14. If someone could take a look at the failing test details in Semaphore that would be great. Looks like we might need to mock an extra object.

maxdeviant (approved):

erin-workos:
Thanks, Marshall!

PR #82, opened 2021-07-20
Clean up test mocking

maxdeviant:
This PR begins the process of cleaning up the mocks in our tests.

Previously we were double mocking our response objects, which was simulating a response object being returned from the WorkOS API, which is not correct.

### DIFF
--- a/workos/audit_trail.py
+++ b/workos/audit_trail.py
@@ -61,7 +61,7 @@ class AuditTrail(object):
             "idempotency-key": idempotency_key,
         }
 
-        self.request_helper.request(
+        response = self.request_helper.request(
             EVENTS_PATH,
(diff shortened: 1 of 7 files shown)


workos/audit_trail.py
```diff
@@ -61,15 +61,15 @@ def create_event(self, event, idempotency_key=None):
...
+        response = self.request_helper.request(
             EVENTS_PATH,
             method=REQUEST_METHOD_POST,
             params=event,
             headers=headers,
             token=workos.api_key,
         )
 
-        return True
+        return response["success"]
```
maxdeviant, line 72: Updated this to make the SDK use the actual response from the API.

marktran (approved):

PR #83, opened 2021-08-11
Add get_profile method and unit test

sheldonvaughn:
PR to add the get_profile method and a corresponding unit test to the python SDK.

### DIFF
--- a/workos/sso.py
+++ b/workos/sso.py
@@ -90,24 +90,22 @@ class SSO(object):
     def get_profile(self, accessToken):
         """
         Verify that SSO has been completed successfully and retrieve the identity of the user.
-        
+
         Args:
             accessToken (str): the token used to authenticate the API call
-        
+
         Returns:
             WorkOSProfile
         """
 
-        params = {
-            "access_token": accessToken
-        }
+        params = {"access_token": accessToken}
 
         response = self.request_helper.request(
             PROFILE_PATH, method=REQUEST_METHOD_POST, params=params
         )
 
         return WorkOSProfile.construct_from_response(response)
-        
+
     def get_profile_and_token(self, code):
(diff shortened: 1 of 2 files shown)


workos/sso.py
```diff
@@ -86,6 +87,25 @@ def get_authorization_url(
...
+        Verify that SSO has been completed successfully and retrieve the identity of the user.
+
+        Args:
+            accessToken (str): the token used to authenticate the API call
+
+        Returns:
+            WorkOSProfile
+        """
+
+        params = {"access_token": accessToken}
```
maxdeviant, line 101: I don't know that `params` is going to do the right thing here. I think this will add an `access_token` query parameter, but we need to set the access token in the `Authorzation: Bearer <ACCESS_TOKEN>` header.

sheldonvaughn: Thanks @maxdeviant  - I see what you're saying. The `request` function has an option for a token

```
if token:
            headers["Authorization"] = "Bearer {}".format(token)
```

 so I will pass the token in there and test the behavior.

workos/sso.py
```diff
@@ -86,6 +87,25 @@ def get_authorization_url(
...
+            accessToken (str): the token used to authenticate the API call
+
+        Returns:
+            WorkOSProfile
+        """
+
+        token = {"access_token": accessToken}
+
+        response = self.request_helper.request(
+            PROFILE_PATH, method=REQUEST_METHOD_GET, token=token
```
maxdeviant, line 104: I think this should just be `token=accessToken` rather than passing in a dictionary.

sheldonvaughn: Fixed, thanks!

maxdeviant (approved):

PR #80, opened 2021-06-16
Fix ServerException for successful DELETE operations

com4:
The connection and organization DELETE endpoints don't return JSON so a `ServerException` is raised on successful operations. (I don't know if this is the case for `DELETE directories/:id` as I have nothing to test against.) 

To reproduce the error with tests I updated the `MockResponse` to deserialize the provided response and updated all tests which used it to provide a serialized string.

To fix the issue this checks the Content-Type header to see if it's JSON before trying to deserialize the response. 

Exception:
```
Traceback (most recent call last):
  <snip>
  File "/app/utils/accounts.py", line 1008, in delete
    sso.workos_client.organizations.delete_organization(
  File "/usr/local/lib/python3.8/site-packages/workos/organizations.py", line 112, in delete_organization
    return self.request_helper.request(
  File "/usr/local/lib/python3.8/site-packages/workos/utils/request.py", line 80, in request
    raise ServerException(response)
workos.exceptions.ServerException: (message=No message, request_id=f801bae5-f482-42fa-96c1-a44c5592d1ee)
```

```
DELETE https://api.workos.com/organizations/org_1a2b3c4d
----------
response.headers:
{'Access-Control-Allow-Credentials': 'true',
 'Connection': 'keep-alive',
 'Content-Length': '8',
 'Content-Security-Policy': "default-src 'self';base-uri "
                            "'self';block-all-mixed-content;font-src 'self' "
                            "https: data:;frame-ancestors 'self';img-src "
                            "'self' data:;object-src 'none';script-src "
                            "'self';script-src-attr 'none';style-src 'self' "
                            "https: 'unsafe-inline';upgrade-insecure-requests",
 'Content-Type': 'text/plain; charset=utf-8',
 'Date': 'Wed, 16 Jun 2021 20:08:59 GMT',
 'Etag': 'W/"8-YaBXLEiT7zQxEyDYTILfiL6oPhE"',
 'Expect-Ct': 'max-age=0',
 'Referrer-Policy': 'no-referrer',
 'Server': 'Cowboy',
 'Strict-Transport-Security': 'max-age=15552000; includeSubDomains',
 'Vary': 'Origin, Accept-Encoding',
 'Via': '1.1 vegur',
 'X-Content-Type-Options': 'nosniff',
 'X-Dns-Prefetch-Control': 'off',
 'X-Download-Options': 'noopen',
 'X-Frame-Options': 'SAMEORIGIN',
 'X-Permitted-Cross-Domain-Policies': 'none',
[...]

PR #84, opened 2021-09-10
Only attempt to parse JSON for JSON responses

maxdeviant:
This PR fixes an issue where we were trying to parse the response body as JSON for API responses that did not return JSON.

We now only parse the response body as JSON if it has a `Content-Type` of `application/json`.

Closes #80.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,8 +10,18 @@ class MockResponse(object):
         self.status_code = status_code
         self.headers = {} if headers is None else headers
 
+        if "content-type" not in self.headers:
+            self.headers["content-type"] = "application/json"
+
     def json(self):
         return self.response_dict
+
+
+class MockRawResponse(object):
+    def __init__(self, content, status_code, headers=None):
+        self.content = content
+        self.status_code = status_code
+        self.headers = {} if headers is None else headers
 
 
 @pytest.fixture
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -10,10 +10,20 @@ def __init__(self, response_dict, status_code, headers=None):
...
+
     def json(self):
         return self.response_dict
 
 
+class MockRawResponse(object):
+    def __init__(self, content, status_code, headers=None):
+        self.content = content
+        self.status_code = status_code
+        self.headers = {} if headers is None else headers
```
mthadley, line 24: Should we move this `{}` default up to the parameter declaration in these mock helpers?

```python
class MockRawResponse(object):
    def __init__(self, content, status_code, headers={}):
        self.content = content
        self.status_code = status_code
        self.headers = headers
```

Not sure if there are any places in the tests that explicitly pass `headers=None` though.

workos/utils/request.py
```diff
@@ -67,10 +67,16 @@ def request(
             response = request_fn(url, headers=headers, json=params)
 
         response_json = None
-        try:
-            response_json = response.json()
-        except ValueError:
-            raise ServerException(response)
+        content_type = (
+            response.headers.get("content-type")
+            if response.headers is not None
```
mthadley, line 72: This null check is reasonable, but do we know when `response.headers` is possibly `None`?

I'm not familiar with the API, so I could see it possibly always defaulting to an empty dictionary. But maybe it is `None` in the case of client errors when there is no response at all?

mthadley (approved):

tests/conftest.py
maxdeviant: Yea, that seems like a good simplification.

I can open up a follow-up PR.

workos/utils/request.py
maxdeviant: I think I added this check after I had seen it in my test cases. Not sure if it is because we have a specific instance where we're setting the headers to `None`.

PR #87, opened 2021-09-15
Handle `Content-Type` headers that contain an encoding

mthadley:
The [`Content-Type`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) header is frequently sent like `application/json`, [but it can also contain extra "subtypes" like `; charset=utf8`](https://www.w3.org/Protocols/rfc1341/4_Content-Type.html). This leads to an issue with our response handling, as it compares the entire header value (like `application/json; charset=urf8`) against `application/json` to determine if the response was JSON.

These changes make that check less exact, and instead we only look to see if the header contains `application/json`, and then attempt to deserialize JSON.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,7 +11,7 @@ class MockResponse(object):
         self.headers = {} if headers is None else headers
 
         if "content-type" not in self.headers:
-            self.headers["content-type"] = "application/json; charset=utf-8"
+            self.headers["content-type"] = "application/json"
 
     def json(self):
         return self.response_dict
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -11,7 +11,7 @@ def __init__(self, response_dict, status_code, headers=None):
         self.headers = {} if headers is None else headers
 
         if "content-type" not in self.headers:
-            self.headers["content-type"] = "application/json"
+            self.headers["content-type"] = "application/json; charset=utf-8"
```
maxdeviant, line 14: Could we leave this as-is and instead add two tests that set the `Content-Type` to `application/json` and `application/json; charset=utf-8` to ensure that both parse the JSON response as expected?

maxdeviant (approved):

Issue #102, opened 2021-11-25
KeyError: "first_name" when calling `workos.sso.SSO.get_profile_and_token`


Genetical:
WorkOS Version: `1.6.0`
Django Version: `3.2.9`
Python Version: `3.9.9`

Successful passwordless session (magic link auth) flows are failing when calling `get_profile_and_token` because the API is not returning `first_name` and `last_name`. This means that during the iteration of `cls.OBJECT_FIELDS` in `workos.resources.base.WorkOSBaseResource.construct_from_response`, the `setattr` function is attempting to access a field in the response dictionary that doesn't exist, hence the `KeyError`.

This is not part of a provider (e.g: Google SSO) SSO process and as such, no `first_name` or `last_name` should be expected as WorkOS would not know these values.

Below is the traceback along with relevant local variables.
**Traceback:**
```python
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/usr/local/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "<... SNIP ...>", line 22, in get
    auth_service.login_user_with_code(code_str=login_code_str, request=request)
  File "/opt/project/authentication/auth_service.py", line 64, in login_user_with_code
    profile_and_token_instance = workos_client.sso.get_profile_and_token(code_str)
  File "/usr/local/lib/python3.9/site-packages/workos/sso.py", line 132, in get_profile_and_token
    return WorkOSProfileAndToken.construct_from_response(response)
  File "/usr/local/lib/python3.9/site-packages/workos/resources/sso.py", line 41, in construct_from_response
    profile_and_token.profile = WorkOSProfile.construct_from_response(
  File "/usr/local/lib/python3.9/site-packages/workos/resources/base.py", line 22, in construct_from_response
    setattr(obj, field, response[field])
```

**Local Vars**
```python
response = {
[...]

PR #103, opened 2021-11-25
Use `None` for fields that do not exist

maxdeviant:
This PR fixes an issue where fields not present in a JSON response would result in a `KeyError` being thrown (as described in #102).

Now, if the key is not present in the response we use `None` as the value instead.

### DIFF
--- a/workos/resources/base.py
+++ b/workos/resources/base.py
@@ -19,7 +19,7 @@ class WorkOSBaseResource(object):
         """
         obj = cls()
         for field in cls.OBJECT_FIELDS:
-            setattr(obj, field, response[field])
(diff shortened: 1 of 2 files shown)


workos/resources/base.py
```diff
@@ -19,7 +19,7 @@ def construct_from_response(cls, response):
         """
         obj = cls()
         for field in cls.OBJECT_FIELDS:
-            setattr(obj, field, response[field])
+            setattr(obj, field, response[field] if field in response else None)
```
Genetical, line 22: `response[field] if field in response else None` can be simplified to `response.get(field)` which will default to `None`.

mthadley (approved):

PR #107, opened 2021-12-17
Create a requirement for RedirectURI in GetAuthorizationURL function

sheldonvaughn:
Currently if customers don't provide a redirectURI in the getAuthorizationURI function they do not get an error from the SDK, but will receive an error in the browser. We can improve the user experience by raising this error in the SDK instead of the browser.

### DIFF
--- a/workos/sso.py
+++ b/workos/sso.py
@@ -102,12 +102,10 @@ class SSO(object):
 
         if state is not None:
             params["state"] = state
-        
+
         if redirect_uri is None:
-            raise ValueError(
-               "Incomplete arguments. Need to specify a Redirect URI"
-            )
-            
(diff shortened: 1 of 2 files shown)


workos/sso.py
```diff
@@ -103,6 +103,9 @@ def get_authorization_url(
         if state is not None:
             params["state"] = state
 
+        if redirect_uri is None:
+            raise ValueError("Incomplete arguments. Need to specify a Redirect URI")
```
mthadley, line 107: We should probably reference the actual parameter name, like we do for the other `ValueError` cases we raise:

```suggestion
            raise ValueError("Incomplete arguments. Need to specify a 'redirect_uri'.")
```

zenorocha:
Can we get this merged?

mthadley (approved):

PR #108, opened 2021-12-20
Add Semaphore promotion for PyPI publishing

robframpton:


maxdeviant:
This is great!

How does it do authentication with PyPI? I didn't see anything in the Semaphore config that looked like an authentication step.

robframpton:
@maxdeviant, great question! The CLI can be authenticated with an implicit environment variable which is what's happening here. 

Docs https://twine.readthedocs.io/en/stable/#environment-variables

maxdeviant:
> @maxdeviant, great question! The CLI can be authenticated with an environment variables which is what's happening here.
> 
> Docs https://twine.readthedocs.io/en/stable/#environment-variables

Ah yes! I now see you're pulling in the `python-pypi` secret.

maxdeviant (approved):

PR #111, opened 2022-02-23
Add error if connection_type is not string

AshGodfrey:
This is in response to the issue raised here: 
https://github.com/workos-inc/workos-python/issues/110

From Marshall: 
> I think we should be safe to add a runtime error if the user passes a string value that does not cast to an enum value, it's fairly safe to assume that moving this validation up into the SDK won't break any existing callers.

This change throws an error if the passed `connection_type` is of `ConnectionType` type rather than string (which the current behavior requires)

FWIW, I'm not strictly certain that this actually requires:  `params["connection_type"] = connection_type` as this is already occurring.

### DIFF
--- a/workos/sso.py
+++ b/workos/sso.py
@@ -204,8 +204,10 @@ class SSO(object):
         }
 
          if connection_type is not None:
-            if isinstance(connection_type, ConnectionType):
+             if isinstance(connection_type, ConnectionType):
                 raise ValueError("'connection_type' must be of type string")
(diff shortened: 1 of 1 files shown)


maxdeviant:
I'm not sure preventing callers from passing the `ConnectionType` is the way we want to go. I think we want to push consumers towards using the enum values.

I opened up #112 as an alternative approach that supports both strings and enum values for the `connection_type` parameter.

PR #112, opened 2022-02-23
Make `connection_type` for `list_connections` accept enum values

### LINKED ISSUE #110: list_connections says it takes a ConnectionType, takes a str
Hello! 

The docs state that this function takes a ConnectionType, so we assumed it did, but it actually takes in a str. I've included the error below that we were seeing on our application.

It would be great to have so error handling as you do here: https://github.com/workos-inc/workos-python/blob/8dae7b1dc8c31db179e7c0e012faeb1d7fd55179/workos/sso.py#L84-L87

```
>>> workos.client.sso.list_connections(domain="foo-corp.com", connection_type=ConnectionType.OktaSAML)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/app/.heroku/python/lib/python3.8/site-packages/workos/sso.py", line 202, in list_connections
    return self.request_helper.request(
  File "/app/.heroku/python/lib/python3.8/site-packages/workos/utils/request.py", line 89, in request
    raise BadRequestException(
workos.exceptions.BadRequestException: (message=Validation failed, request_id=...)

(message=Validation failed, request_id=45831dcd-650a-4ac2-bd40-eb96717420f7)
```

```
{
  "date": "2022-02-16T00:12:50.856Z",
  "id": "...",
  "method": "GET",
  "path": "/connections",
  "status": 422,
  "source": "WorkOS Python/1.7.0",
  "request_query": "{\"connection_type\":\"ConnectionType.OktaSAML\",\"domain\":\"foo-corp.com\",\"limit\":\"10\"}",
  "request_body": null,
  "response_body": {
    "errors": [
      {
        "code": "connection_type must be a valid enum value",
        "field": "connection_type"
      }
    ],
    "message": "Validation failed"
  }
}
```

Running the 1.7.0 version on Python 3.8.

maxdeviant:
This PR updates the handling of the `connection_type` parameter for the `sso.list_connections` method to support passing a `ConnectionType` enum value.

To retain backwards compatibility, we attempt to coerce any string values to their corresponding `ConnectionType` values. If given an invalid string, an exception will be thrown. This should not be a breaking change, as we already validate the connection types on the API side. All we're doing is moving the validation up into the SDK.

Fixes #110.

Closes #111.

### DIFF
--- a/tests/test_sso.py
+++ b/tests/test_sso.py
@@ -354,6 +354,42 @@ class TestSSO(object):
 
         assert connections_response["data"] == mock_connections["data"]
 
+    def test_list_connections_with_connection_type_as_invalid_string(
+        self, setup_with_client_id, mock_connections, mock_request_method
+    ):
+        mock_request_method("get", mock_connections, 200)
+
+        try:
+            self.sso.list_connections(connection_type="UnknownSAML")
+        except Exception as ex:
+            assert str(ex) == "'connection_type' must be a member of ConnectionType"
+
+    def test_list_connections_with_connection_type_as_string(
+        self, setup_with_client_id, mock_connections, capture_and_mock_request
+    ):
+        request_args, request_kwargs = capture_and_mock_request(
+            "get", mock_connections, 200
+        )
+
+        connections_response = self.sso.list_connections(connection_type="GenericSAML")
+
+        request_params = request_kwargs["params"]
+        assert request_params["connection_type"] == "GenericSAML"
+
+    def test_list_connections_with_connection_type_as_enum(
+        self, setup_with_client_id, mock_connections, capture_and_mock_request
+    ):
+        request_args, request_kwargs = capture_and_mock_request(
+            "get", mock_connections, 200
+        )
+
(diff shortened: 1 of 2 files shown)


tests/test_sso.py
```diff
@@ -354,6 +354,42 @@ def test_list_connections(
...
 
+    def test_list_connections_with_connection_type_as_invalid_string(
+        self, setup_with_client_id, mock_connections, mock_request_method
+    ):
+        mock_request_method("get", mock_connections, 200)
+
+        try:
+            self.sso.list_connections(connection_type="UnknownSAML")
+        except Exception as ex:
+            assert str(ex) == "'connection_type' must be a member of ConnectionType"
```
mthadley, line 365: Can we use `pytest.raises`, and also assert on a more specific error type?

```suggestion
        with pytest.raises(
            ValueError, match="'connection_type' must be a member of ConnectionType"
        ):
            self.sso.list_connections(connection_type="UnknownSAML")
```

mthadley (approved):

PR #114, opened 2022-02-28
Add Organization parameter to List Directories

awolfden:
Add Organization parameter to List Organizations

### DIFF
--- a/workos/directory_sync.py
+++ b/workos/directory_sync.py
@@ -123,22 +123,22 @@ class DirectorySync(object):
 
         Args:
             domain (str): Domain of a Directory. (Optional)
+            organization: ID of an Organization (Optional)
             search (str): Searchable text for a Directory. (Optional)
             limit (int): Maximum number of records to return. (Optional)
             before (str): Pagination cursor to receive records before a provided Directory ID. (Optional)
             after (str): Pagination cursor to receive records after a provided Directory ID. (Optional)
-            organization: ID of an Organization (Optional)
 
         Returns:
             dict: Directories response from WorkOS.
         """
         params = {
             "domain": domain,
+            "organization_id": organization,
             "search": search,
             "limit": limit,
             "before": before,
             "after": after,
-            "organization_id": organization,
         }
         return self.request_helper.request(
             "directories",
--- a/workos/directory_sync.py
+++ b/workos/directory_sync.py
@@ -111,12 +111,13 @@ class DirectorySync(object):
         )
 
     def list_directories(
-        self, domain=None, search=None, limit=RESPONSE_LIMIT, before=None, after=None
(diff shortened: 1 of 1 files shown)


workos/directory_sync.py
```diff
@@ -111,12 +111,13 @@ def get_group(self, group):
         )
 
     def list_directories(
-        self, domain=None, search=None, limit=RESPONSE_LIMIT, before=None, after=None
+        self, domain=None, organization=None, search=None, limit=RESPONSE_LIMIT, before=None, after=None
```
maxdeviant, line 114: I think `organization` should technically be added to the end, as if a caller is calling `list_directories` with positional arguments (instead of named arguments) then shifting the existing arguments would be a breaking change.

I think Python has ways to force callers to use named arguments, so that might be something we want to look into at some point.

awolfden: For sure, and I just switched organization to be the last param

workos/directory_sync.py
```diff
@@ -111,12 +111,19 @@ def get_group(self, group):
         )
 
     def list_directories(
-        self, domain=None, search=None, limit=RESPONSE_LIMIT, before=None, after=None
+        self,
+        domain=None,
+        organization=None,
```
maxdeviant, line 116: It looks like `organization` is still where it was before, the formatting just changed 🤔

awolfden: Oh man I made the change in the example app where i was testing things out, not the sdk. 🤦🏽   This is fixed now!

workos/directory_sync.py
```diff
@@ -121,6 +127,7 @@ def list_directories(
             limit (int): Maximum number of records to return. (Optional)
             before (str): Pagination cursor to receive records before a provided Directory ID. (Optional)
             after (str): Pagination cursor to receive records after a provided Directory ID. (Optional)
+            organization: ID of an Organization (Optional)
```
maxdeviant, line 130: I think in the docstring we can keep it in the same position, since it makes more sense having it first (and it doesn't have any effect on compatibility).

workos/directory_sync.py
```diff
@@ -131,6 +138,7 @@ def list_directories(
             "limit": limit,
             "before": before,
             "after": after,
+            "organization_id": organization,
```
maxdeviant, line 141: Same thing here. I think we can leave this up by `domain`, since it fits there better since they're both filters.

workos/directory_sync.py
awolfden: 👍🏽

workos/directory_sync.py
awolfden: 👍🏽

maxdeviant (approved):

PR #116, opened 2022-03-08
Add MFA support

awolfden:
Add MFA to Python SDK


robframpton:
Method signatures look good to me, would love someone with more Python prowess to give a look as well 👀

workos/mfa.py
```diff
@@ -0,0 +1,136 @@
...
+    REQUEST_METHOD_POST,
+)
+from workos.utils.validation import MFA_MODULE, validate_settings
+
+ENROLL_PATH = "auth/factors/enroll"
+CHALLENGE_PATH = "auth/factors/challenge"
+VERIFY_PATH = "auth/factors/verify"
+
+
+class MFA(object):
```
maxdeviant, line 13: nit: Can we change the casing to `Mfa`?

tests/test_mfa.py
```diff
@@ -0,0 +1,198 @@
+from workos.mfa import MFA
+import pytest
+
+
+class TestWebhooks(object):
```
maxdeviant, line 5: Looks like this name was copied over:

```suggestion
class TestMfa(object):
```

awolfden: ✔️

workos/mfa.py
awolfden: ✔️

pliao28:
@awolfden Can you change the version for a pre-release version (believe it would be 1.11.0.a1 https://pythonpackaging.info/07-Package-Release.html) 
@robframpton could you check if the semaphore publish run needs any adjustment

pliao28:
@awolfden  @robframpton , published can be retrieved by `pip install 'workos==1.11.0.a1' `

workos/__about__.py
```diff
@@ -12,7 +12,7 @@
 
 __package_url__ = "https://github.com/workos-inc/workos-python"
 
-__version__ = "1.10.0"
+__version__ = "1.11.0.a1"
```
robframpton, line 15: Synced with the eng team today. We're going to be publishing MFA changes to the stable channels so we don't need to worry about prerelease suffixes.
```suggestion
__version__ = "1.10.0"
```

robframpton (approved):
@awolfden once the `__version__` change I suggested is committed this will be good to go.

PR #124, opened 2022-05-31
Add order to pagination params

pliao28:


workos/directory_sync.py
```diff
@@ -7,6 +7,7 @@
 from workos.utils.validation import DIRECTORY_SYNC_MODULE, validate_settings
 
 RESPONSE_LIMIT = 10
+ORDER_DEFAULT = "normal"
```
hadihallak, line 10: Just to be consistent across SDKs I think we should make the default whatever the equivalent for undefined in python

hadihallak (commented):
I touched python once or twice in my whole life so I'm not sure if we are able to type values this way but it would be nice if we could type the available values somewhere `asc` | `desc`

hadihallak:
not sure how optional args work in python but I think we might need to move the order param to be the last param like @sheldonvaughn suggested in the SDK PR for PHP to make sure that this isn't a breaking change.

pliao28:
> not sure how optional args work in python but I think we might need to move the order param to be the last param like @sheldonvaughn suggested in the SDK PR for PHP to make sure that this isn't a breaking change.

I don't think param ordering conventions matter much for Python, as they use assignment operators, but I will change the ordering across the board just to match for consistency. Originally I was thinking of setting it after limit because that would make the most sense when thinking about pagination, where you would set limit and order first

hadihallak:
Yeah I totally agree. I also think that the new planed methods for pagination would probably render these params obsolete so I totally agree with you.

PR #125, opened 2022-06-06
Add class defining Directory Group dict 

awolfden:
Adds a class to define the properties returned from a Directory Group as WorkOSDirectoryGroup. 

Updates Enum list with current list of connection types.


pliao28 (approved):
lgtm!

PR #126, opened 2022-06-07
Add MFA WorkOSChallenge class to make challenge response fields non-nullable

awolfden:
Add MFA WorkOSChallenge class to make challenge response fields non-nullable


pliao28 (approved):
lgtm

PR #128, opened 2022-06-29
Allow updating organization's `allow_profiles_outside_organization` flag

GeyseR:
probably the docs page should be updated as well


maxdeviant (approved):
Thanks for the PR!

PR #131, opened 2022-07-05
Update endpoints used for MFA operations

maxdeviant:
This PR updates the endpoints used for two of the MFA operations:

- `challenge_factor`
- `verify_challenge`

The `verify_factor` method is now deprecated and has been replaced with `verify_challenge`. `verify_factor` will be removed in the next major version.

Resolves SDK-547.

### DIFF
--- a/workos/mfa.py
+++ b/workos/mfa.py
@@ -119,6 +119,8 @@ class Mfa(object):
         """
         Verifies the one time password provided by the end-user.
 
+        Deprecated: Please use `verify_challenge` instead.
+
         Kwargs:
             authentication_challenge_id (str) - The ID of the authentication challenge that provided the user the verification code.
             code (str) - The verification code sent to and provided by the end user.
--- a/workos/mfa.py
+++ b/workos/mfa.py
@@ -1,3 +1,5 @@
+from warnings import warn
+
 import workos
 from workos.utils.request import (
     RequestHelper,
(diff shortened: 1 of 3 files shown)


workos/mfa.py
```diff
@@ -127,6 +126,11 @@ def verify_factor(
         Returns: Dict containing the  challenge factor details.
         """
 
+        warn(
+            "'verify_factor' is deprecated. Please use 'verify_challenge' instead.",
+            DeprecationWarning,
+        )
+
         params = {
```
mthadley, line 134: Given the differences between `verify_factor` and `verify_challenge` are...

* One has a `warn` since it is deprecated
* The other uses the new `auth/challenges/{challenge_id}/verify` endpoint, but is otherwise backwards compatible

Can we define `veryify_factor` in terms of `verify_challenge` until it is finally removed?

```py
    def verify_factor(
        self,
        authentication_challenge_id=None,
        code=None,
    ):
        """
        Verifies the one time password provided by the end-user.

        Kwargs:
            authentication_challenge_id (str) - The ID of the authentication challenge that provided the user the verification code.
            code (str) - The verification code sent to and provided by the end user.

        Returns: Dict containing the  challenge factor details.
        """

        warn(
            "'verify_factor' is deprecated. Please use 'verify_challenge' instead.",
            DeprecationWarning,
        )

        return self.verify_challenge(
           authentication_challenge_id=autentication_challenge_id, code=code
        )
```

workos/mfa.py
```diff
@@ -127,6 +126,11 @@ def verify_factor(
         Returns: Dict containing the  challenge factor details.
```
mthadley, line 126: Should we also update the Docstring for `verify_factor` to mention that is now deprecated?

workos/mfa.py
maxdeviant: That's a good point.

I updated `verify_factor` to be implemented in terms of `verify_challenge`.

workos/mfa.py
maxdeviant: Done!

mthadley (approved):
Looking good!

PR #135, opened 2022-07-19
Refactor methods to use WorkOS object classes and update tests

awolfden:
Refactor methods to use WorkOS object classes and update tests


logangingerich (approved):
superb

PR #136, opened 2022-07-20
Update SDK version 1.14.0

awolfden:
Update SDK version to 1.14.0


maxdeviant:
Could we get in the habit of using the SDK version as the PR title?

Otherwise it makes it much harder to track versions in Git history.

logangingerich (approved):

PR #139, opened 2022-08-18
Add Audit Logs to the Python SDK

jonatascastro12:
Add AuditLogs calls to the SDK:
-  `create_event`
-  `create_export`
-  `get_export`

Also made a few tweaks to expose `code` and `errors` payload in exceptions.

### DIFF
--- a/workos/audit_logs.py
+++ b/workos/audit_logs.py
@@ -1,0 +1,66 @@
+import workos
+from workos.utils.pagiantion_order import Order
+from workos.exceptions import ConfigurationException
+from workos.resources.event import WorkOSEvent
+from workos.utils.request import RequestHelper, REQUEST_METHOD_GET, REQUEST_METHOD_POST
+from workos.utils.validation import AUDIT_TRAIL_MODULE, validate_settings
+
+EVENTS_PATH = "audit_logs/events"
+EXPORTS_PATH = "audit_logs/exports"
+
+
+class AuditLogs(object):
+    """Offers methods through the WorkOS Audit Logs service."""
+
+    @validate_settings(AUDIT_TRAIL_MODULE)
+    def __init__(self):
+        pass
+
+    @property
+    def request_helper(self):
+        if not getattr(self, "_request_helper", None):
+            self._request_helper = RequestHelper()
+        return self._request_helper
+
+    def create_event(self, organization_id, event, idempotency_key=None):
+        """Create an Audit Logs event.
+
+        Args:
+            organization_id (str) - Organization's unique identifier
+            event (dict) - An event object
+                event[action] (string) - The event action
+                event[version] (int) - The version of event
+                event[occurred_at] (datetime) - ISO-8601 datetime of when an event occurred
+                event[actor] (dict) - Describes the entity that generated the event
+                    event[actor][id] (str)
+                    event[actor][name] (str)
+                    event[actor][type] (str)
+                    event[actor][metadata] (dict)
+                event[targets] (list[dict])
+                event[context] (dict)
+                    event[context][location] (str)
+                    event[context][user_agent] (str)
+                event[metadata] (dict)
+            idempotency_key (str) - An idempotency key
+
+        Returns:
+            boolean: Returns True
+        """
+        payload = {
+            "organization_id": organization_id,
+            "event": event
+        }
+
+        headers = {}
+        if idempotency_key:
+            headers["idempotency-key"] = idempotency_key
+
+        response = self.request_helper.request(
+            EVENTS_PATH,
+            method=REQUEST_METHOD_POST,
+            params=payload,
+            headers=headers,
+            token=workos.api_key,
+        )
+
+        return response["success"]
--- a/workos/audit_logs.py
+++ b/workos/audit_logs.py
@@ -36,20 +36,17 @@ class AuditLogs(object):
                     event[actor][name] (str)
                     event[actor][type] (str)
                     event[actor][metadata] (dict)
-                event[targets] (list[dict])
-                event[context] (dict)
+                event[targets] (list[dict]) - List of event targets
+                event[context] (dict) - Attributes of event context
                     event[context][location] (str)
                     event[context][user_agent] (str)
-                event[metadata] (dict)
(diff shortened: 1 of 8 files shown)


workos/audit_logs.py
```diff
@@ -0,0 +1,126 @@
...
+                    event[actor][id] (str)
+                    event[actor][name] (str)
+                    event[actor][type] (str)
+                    event[actor][metadata] (dict)
+                event[targets] (list[dict]) - List of event targets
+                event[context] (dict) - Attributes of event context
+                    event[context][location] (str)
+                    event[context][user_agent] (str)
+                event[metadata] (dict) - Extra metadata
+            idempotency_key (str) - Optional idempotency key
```
jonatascastro12, line 45: @robframpton Could take a look at the descriptions of the fields on this docstring?

workos/audit_logs.py
```diff
@@ -0,0 +1,126 @@
...
+    ):
+        """Trigger the creation of an export of audit logs.
+
+        Args:
+            organization (str) - Organization's unique identifier
+            range_start (str) - Start date of the range filter
+            range_end (str) - End date of the range filter
+            actions (list) - Optional list of actions to filter
+            actors (list) - Optional list of actors to filter
+            targets (list) - Optional list of targets to filter
```
jonatascastro12, line 83: Here as well. @robframpton

workos/__about__.py
```diff
@@ -12,7 +12,7 @@
 
 __package_url__ = "https://github.com/workos-inc/workos-python"
 
-__version__ = "1.14.0"
+__version__ = "1.15.0"
```
robframpton, line 15: We typically bump the versions in a dedicated branch after feature changes have been merged in.

workos/audit_logs.py
```diff
@@ -0,0 +1,126 @@
...
+        return self._request_helper
+
+    def create_event(self, organization, event, idempotency_key=None):
+        """Create an Audit Logs event.
+
+        Args:
+            organization (str) - Organization's unique identifier
+            event (dict) - An event object
+                event[action] (string) - The event action
+                event[version] (int) - The version of event
```
robframpton, line 33:
```suggestion
                event[version] (int) - The schema version of the event
```

workos/audit_logs.py
```diff
@@ -0,0 +1,126 @@
...
+        Args:
+            organization (str) - Organization's unique identifier
+            range_start (str) - Start date of the range filter
+            range_end (str) - End date of the range filter
+            actions (list) - Optional list of actions to filter
+            actors (list) - Optional list of actors to filter
+            targets (list) - Optional list of targets to filter
+
+        Returns:
+            dict: Object that describes the exported audit logs job
```
robframpton, line 86:
```suggestion
            dict: Object that describes the audit log export
```

workos/audit_logs.py
```diff
@@ -0,0 +1,126 @@
...
+            token=workos.api_key,
+        )
+
+        return WorkOSAuditLogsExport.construct_from_response(response)
+
+    def get_export(self, export_id):
+        """Retrieve an created export.
+
+        Returns:
+            dict: Object that describes the exported event
```
robframpton, line 117:
```suggestion
            dict: Object that describes the audit log export
```

workos/resources/audit_logs_export.py
```diff
@@ -0,0 +1,28 @@
+from workos.resources.base import WorkOSBaseResource
+from workos.resources.event_action import WorkOSEventAction
+
+
+class WorkOSAuditLogsExport(WorkOSBaseResource):
```
robframpton, line 5: This is what we've been calling the entity elsewhere.
```suggestion
class WorkOSAuditLogExport(WorkOSBaseResource):
```

workos/audit_logs.py
```diff
@@ -0,0 +1,126 @@
...
+        actions=None,
+        actors=None,
+        targets=None,
+    ):
+        """Trigger the creation of an export of audit logs.
+
+        Args:
+            organization (str) - Organization's unique identifier
+            range_start (str) - Start date of the range filter
+            range_end (str) - End date of the range filter
```
robframpton, line 80:
```suggestion
            range_start (str) - Start date of the date range filter
            range_end (str) - End date of the date range filter
```

workos/audit_logs.py
```diff
@@ -0,0 +1,126 @@
...
+
+        response = self.request_helper.request(
+            EVENTS_PATH,
+            method=REQUEST_METHOD_POST,
+            params=payload,
+            headers=headers,
+            token=workos.api_key,
+        )
+
+        return response["success"]
```
robframpton, line 64: Not sure if it's possible in Python, but we've been returning `Void` in the other SDKs as the `success` property is kind of a meaningless value that we return from various endpoints. There is some more context in this Slack thread that lead to this: https://work-os.slack.com/archives/C037Z7GMMAB/p1657562404232559

robframpton:
It looks like we'll need to add some additional error mapping here: https://github.com/workos/workos-python/blob/main/workos/utils/request.py#L87-L96 similarly to what I did for the Kotlin SDK: https://github.com/workos/workos-kotlin/pull/127/files#diff-31e1355895fca164d2b3ce61cd7fc1d1bfbf83c304893e6727fc010a67ad19dcR252-R255

Essentially we just want to make sure the `code` and `errors` properties are propagated.

workos/audit_logs.py
jonatascastro12: So `void` in Python means `return None`. Nothing wrong with that, but I was trying to be consistent with other places in the Python SDK. 

https://github.com/workos/workos-python/blob/78f705670cfb76e8c9888e45a1148fea34ac0763/workos/passwordless.py#L70
https://github.com/workos/workos-python/blob/d5171cad529142b2470b5c69db47516776ac9b41/workos/audit_trail.py#L73

What do you think?

By the way, we don't have Python typing in our SDK (I want to propose that).

robframpton: Yeah it's going against historical convention in the other SDKs as well, I think we'll still want to make the change though to keep the various `createEvent` implementations the same.

PR #140, opened 2022-08-22
Add Python version to the SDK user agent

jonatascastro12:
**BEFORE:**
`"User-Agent": "WorkOS Python/1.14.1"`

**AFTER:**
`"User-Agent": "WorkOS Python/3.10.6 Python SDK/1.14.1"`

### DIFF
--- a/workos/utils/request.py
+++ b/workos/utils/request.py
@@ -1,3 +1,5 @@
+import platform
+
 import requests
 
 import workos
@@ -9,7 +11,9 @@ from workos.exceptions import (
 )
 
 BASE_HEADERS = {
-    "User-Agent": "WorkOS Python/{}".format(workos.__version__),
+    "User-Agent": "WorkOS Python SDK/{} Python/{}".format(
+        workos.__version__, platform.python_version()
+    ),
 }
 
 RESPONSE_TYPE_CODE = "code"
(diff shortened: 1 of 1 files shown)


workos/utils/request.py
```diff
@@ -9,7 +11,9 @@
 )
 
 BASE_HEADERS = {
-    "User-Agent": "WorkOS Python/{}".format(workos.__version__),
+    "User-Agent": "WorkOS Python SDK/{} Python/{}".format(
```
maxdeviant, line 14: It looks like in the Ruby SDK we swap the order here, showing the Python version first and then the SDK version.

Should we keep it consistent?

jonatascastro12: I'm ok with both. No strong opinions. I'll push a commit. 👍

workos/utils/request.py
```diff
@@ -9,7 +11,10 @@
 )
 
 BASE_HEADERS = {
-    "User-Agent": "WorkOS Python/{}".format(workos.__version__),
+    "User-Agent": "Python/{} WorkOS Python SDK/{}".format(
```
maxdeviant, line 14: Could we keep the "WorkOS" first? Again, to stay consistent with the Ruby SDK:

```suggestion
    "User-Agent": "WorkOS Python/{} Python SDK/{}".format(
```

maxdeviant (approved):

PR #141, opened 2022-08-25
Update version to 1.15.0

jonatascastro12:
This should update the SDK to publish the following PRs:
https://github.com/workos/workos-python/pull/139
https://github.com/workos/workos-python/pull/140


erikengervall (approved):
Wonder if we should add this particular step to the Python header in https://www.notion.so/workos/SDK-Release-Process-a52c9c5a71ce48578e7e3e26fc82b386 🤔

PR #142, opened 2022-08-30
Add get_primary_email method to Directory Sync module

awolfden:
Add get_primary_email method to Directory Sync module

### DIFF
--- a/workos/directory_sync.py
+++ b/workos/directory_sync.py
@@ -131,23 +131,6 @@ class DirectorySync(object):
 
         return WorkOSDirectoryUser.construct_from_response(response).to_dict()
 
-    def get_primary_email(self, user):
-        """Gets primary email for a single provisioned Directory User.
-
-        Args:
-            user (str): Directory User unique identifier.
-
-        Returns:
-            str: User's primary email response from WorkOS.
-        """
-        response = self.request_helper.request(
-            "directory_users/{user}".format(user=user),
-            method=REQUEST_METHOD_GET,
-            token=workos.api_key,
-        )
-
-        return WorkOSDirectoryUser.get_primary_email(response)
-
     def get_group(self, group):
         """Gets details for a single provisioned Directory Group.
 
--- a/workos/directory_sync.py
+++ b/workos/directory_sync.py
@@ -145,13 +145,8 @@ class DirectorySync(object):
             method=REQUEST_METHOD_GET,
             token=workos.api_key,
         )
-        primary_email_array = []
-        for i in response["emails"]:
-            if i["primary"]:
-                primary_email_array.append(i["value"])
-        if len(primary_email_array) > 0:
-            return primary_email_array[0]
-        return None
+
+        return WorkOSDirectoryUser.get_primary_email(response)
 
     def get_group(self, group):
         """Gets details for a single provisioned Directory Group.
--- a/workos/directory_sync.py
+++ b/workos/directory_sync.py
@@ -131,6 +131,28 @@ class DirectorySync(object):
 
         return WorkOSDirectoryUser.construct_from_response(response).to_dict()
 
+    def get_primary_email(self, user):
+        """Gets primary email for a single provisioned Directory User.
+
+        Args:
+            user (str): Directory User unique identifier.
+
+        Returns:
+            str: User's primary email response from WorkOS.
+        """
+        response = self.request_helper.request(
+            "directory_users/{user}".format(user=user),
+            method=REQUEST_METHOD_GET,
+            token=workos.api_key,
+        )
+        primary_email_array = []
+        for i in response["emails"]:
+            if i["primary"]:
+                primary_email_array.append(i["value"])
+        if len(primary_email_array) > 0:
+            return primary_email_array[0]
+        return None
+
     def get_group(self, group):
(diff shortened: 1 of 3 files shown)


maxdeviant:
Should we add this as a method on the `WorkOSDirectoryUser` resource instead?

https://github.com/workos/workos-python/blob/59872bf926be24e922ede031166b173f7df792a6/workos/resources/directory_sync.py#L63

workos/directory_sync.py
```diff
@@ -131,6 +131,23 @@ def get_user(self, user):
...
+    def get_primary_email(self, user):
+        """Gets primary email for a single provisioned Directory User.
+
+        Args:
+            user (str): Directory User unique identifier.
+
+        Returns:
+            str: User's primary email response from WorkOS.
+        """
+        response = self.request_helper.request(
```
maxdeviant, line 143: We shouldn't be making an API request here.

The `get_primary_email` method on the `WorkOSDirectoryUser` is an instance method that the developer can call on the directory users returned from the other methods (like `get_user`).

workos/directory_sync.py
```diff
@@ -131,6 +131,23 @@ def get_user(self, user):
 
         return WorkOSDirectoryUser.construct_from_response(response).to_dict()
 
+    def get_primary_email(self, user):
```
maxdeviant, line 134: We don't need this method on the service class itself.

The `get_primary_email` method on the `WorkOSDirectoryUser` is an instance method that the developer can call on the directory users returned from the other methods (like `get_user`).

awolfden: Got it, will work on refactoring this.

workos/resources/directory_sync.py
```diff
@@ -92,3 +97,12 @@ def to_dict(self):
         directory_group = super(WorkOSDirectoryUser, self).to_dict()
 
         return directory_group
+
+    def primary_email(self, user):
```
maxdeviant, line 101: Since this is an instance method we shouldn't take in the `user` here, instead we should return the primary email of the user held in `self`.

workos/resources/directory_sync.py
```diff
@@ -92,3 +97,12 @@ def to_dict(self):
...
+        primary_email_array = []
```
maxdeviant, line 102: I think we could simplify the implementation here using something along the lines of a `.find` (which seems to be curiously absent from Python):

```py
def primary_email(self):
    return next(email for email in user["emails"] if email["primary"], None)
```

Here we're using a list comprehension to filter down the emails to the ones where `primary` is truthy. We then use [`next`](https://docs.python.org/3.7/library/functions.html#next) to return the next item from the iterator. The `None` as the second argument is what we return if there are no primary emails (it will raise a `StopIteration` exception otherwise).

awolfden: Thank you for explaining this, that makes it look much cleaner! 🤌

awolfden: Hey @maxdeviant - I'm having a bit of trouble getting the tests to pass for both Python 2.7 and 3.7. Currently they are only passing for 3.7. 

the Python 2.7 Semaphore run is failing pytest for supplying the primary_email method with a dict instead of an instance of the WorkOSDirectoryUser class. In the SDK, the get_user method returns a dict, not an instance of the class, so it is my assumption that this method should indeed take a dict, not an instance of the class. 

Right now, when I structure the call like the below it returns the primary email as expected.
```primary_email = WorkOSDirectoryUser.primary_email(mock_user_instance)```

Since the SDK returns a dict, calling the method like the below here throws an error regarding a dict not having the method primary_email(). 
```mock_user_instance.primary_email()```

When I adjust the get_user() method to return an instance of the WorkOSDirectoryUser class itself, I get an error referencing the class not being subscriptable when attempting to access the primary_email method. 

I'm not sure how to approach this in order for it to get around this for both Python 2.7 or 3.7 while maintaining the syntax that you had suggested for the method. Do you have any ideas on how to best approach this?

Comment (maxdeviant) [reply workos/resources/directory_sync.py]: If I'm understanding this correctly, it sounds like the issue here is that the `get_user` method is returning a plain dictionary instead of the `WorkOSDirectoryUser` class.

I think for the purposes of this PR we should test the `primary_email` method in isolation based on an instance of the `WorkOSDirectoryUser`.

We'll then eventually want to make `get_user` return a `WorkOSDirectoryUser`. This is a breaking change though, so we'll need to manage that.

Comment (awolfden) [reply workos/resources/directory_sync.py]: Sounds good, I got the tests working now! 👍🏽

**Comment by blairlunceford** [comment]
@maxdeviant does this PR look good to go? Thanks!

File: workos/resources/directory_sync.py [hunk 5]
```diff
@@ -1,4 +1,9 @@
+import workos
 from workos.resources.base import WorkOSBaseResource
+from workos.utils.request import (
```
maxdeviant, line 3: Are these imports necessary? They don't seem to be required for the addition of `primary_email`.

workos/resources/directory_sync.py
```diff
@@ -1,4 +1,9 @@
+import workos
```
maxdeviant, line 1: Is the addition of this import necessary?

maxdeviant (approved):

PR #143, opened 2022-08-31
Add support for idempotency key in create_organization method

awolfden:
Add support for idempotency key in create_organization method

### DIFF
--- a/tests/test_organizations.py
+++ b/tests/test_organizations.py
@@ -114,7 +114,7 @@ class TestOrganizations(object):
             payload, idempotency_key=idempotency_key
         )
 
-        assert request_kwargs["headers"]["Idempotency-Key"] == idempotency_key
+        assert request_kwargs["headers"]["idempotency-key"] == idempotency_key
         assert response["name"] == "Foo Corporation"
 
     def test_update_organization(self, mock_organization_updated, mock_request_method):
--- a/tests/test_organizations.py
+++ b/tests/test_organizations.py
@@ -111,7 +111,7 @@ class TestOrganizations(object):
         _, request_kwargs = capture_and_mock_request("post", payload, 200)
 
         response = self.organizations.create_organization(
-            payload, create_organization_options={"idempotency_key": idempotency_key}
+            payload, idempotency_key=idempotency_key
         )
 
         assert request_kwargs["headers"]["Idempotency-Key"] == idempotency_key
--- a/tests/test_organizations.py
+++ b/tests/test_organizations.py
@@ -1,9 +1,4 @@
-import json
-from requests import Response
-
 import pytest
-
-import workos
 from workos.organizations import Organizations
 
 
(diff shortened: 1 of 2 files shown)


workos/organizations.py
```diff
@@ -88,14 +88,23 @@ def create_organization(self, organization):
                     within the Organization allow profiles that are outside of the Organization's
                     configured User Email Domains. (Optional)
                 organization[domains] (list) - List of domains that belong to the organization
+            create_organization_options (dict) - Options configuration for creating an organization. (Optional)
```
blairlunceford, line 91: Can we model this more after the Audit Logs implementation? https://github.com/workos/workos-python/blob/main/workos/audit_logs.py#L23

Key differences:
- We can just pass an `idempotency_key` parameter with a default value of `None`
- Construct the header like:
```
 headers = {}
        if idempotency_key:
            headers["idempotency-key"] = idempotency_key
```
- pass those `headers` into the request method

tests/test_organizations.py
```diff
@@ -1,9 +1,4 @@
-import json
```
blairlunceford, line 1: Why are these imports deleted?

blairlunceford (changes requested):

tests/test_organizations.py
awolfden: the json import was not being used so I removed it to clean up the file. If this is being used somewhere that you see I'm happy to add it back in.

workos/organizations.py
awolfden: Updated ✅

tests/test_organizations.py
blairlunceford: Sounds good 👍

workos/organizations.py
```diff
@@ -88,14 +88,21 @@ def create_organization(self, organization):
                     within the Organization allow profiles that are outside of the Organization's
                     configured User Email Domains. (Optional)
                 organization[domains] (list) - List of domains that belong to the organization
+            idempotency_key (str) - Idempotency key for creating an organization. (Optional)
+                idempotency_key (str) - A custom key to ensure uniqueness when creating a new organization.
```
blairlunceford, line 92: Don't think we need this line anymore

workos/organizations.py
```diff
@@ -88,14 +88,21 @@ def create_organization(self, organization):
...
 
         Returns:
             dict: Created Organization response from WorkOS.
         """
+        headers = {"Idempotency-Key": None}
+        if idempotency_key:
+            headers["Idempotency-Key"] = idempotency_key
```
blairlunceford, line 99: Should be: `headers["idempotency-key"] = idempotency_key`

blairlunceford (changes requested):

awolfden:
Hey @blairlunceford I think I covered everything that you flagged, but let me know if you see anything else that needs to get updated!

blairlunceford (approved):
Looks good 🎉

PR #152, opened 2022-10-10
Add support for delete_factor and get_factor methods

awolfden:
- Add support for delete factor and get_factor methods
- Add tests for delete_factor and get_factor methods


tests/test_mfa.py
```diff
@@ -169,34 +180,34 @@ def test_challenge_success(
         )
         assert challenge_factor == mock_challenge_factor_response
 
-    def test_verify_factor_no_id(self, mock_verify_challenge_payload):
+    def test_verify_challenge_no_id(self, mock_verify_challenge_payload):
         with pytest.raises(ValueError) as err:
-            self.mfa.verify_factor(
+            self.mfa.verify_challenge(
```
maxdeviant, line 184: We want to keep the `verify_factor` tests in place to ensure the `verify_factor` method is still under test.

We already have test coverage for the `verify_challenge` method.

We can remove these tests when we remove the `verify_factor` method in the next major version.

workos/mfa.py
```diff
@@ -88,6 +88,30 @@ def enroll_factor(
...
+
+        Returns: Does not provide a return.
+        """
+
+        if id is None:
+            raise ValueError("Incomplete arguments. Need to specify a factor id.")
+
+        response = self.request_helper.request(
+            "auth/factors/{id}",
+            method=REQUEST_METHOD_POST,
```
maxdeviant, line 109: Should this be a DELETE?

```suggestion
            method=REQUEST_METHOD_DELETE,
```

tests/test_mfa.py
awolfden: That makes sense! I just removed those changes. ☑️

workos/mfa.py
```diff
@@ -88,6 +89,59 @@ def enroll_factor(
 
         return WorkOSAuthenticationFactorSms.construct_from_response(response).to_dict()
 
+    def get_factor(
+        self,
+        id=None,
+    ):
+        """
+        Returns an authorization factor from it's ID.
```
maxdeviant, line 97: Possessive "its" does not use an apostrophe:

```suggestion
        Returns an authorization factor from its ID.
```

workos/mfa.py
```diff
@@ -88,6 +89,59 @@ def enroll_factor(
...
+
+        Kwargs:
+            id (str) - The id of the factor to be obtained.
```
maxdeviant, line 100: nit: casing:

```suggestion
            id (str) - The ID of the factor to be obtained.
```

workos/mfa.py
```diff
@@ -88,6 +89,59 @@ def enroll_factor(
...
+        Deletes an MFA authorization factor.
+
+        Kwargs:
+            id (str) - The id of the authorization factor to be deleted.
+
+        Returns: Does not provide a return.
+        """
+
+        if id is None:
+            raise ValueError("Incomplete arguments. Need to specify a factor id.")
```
maxdeviant, line 135:
```suggestion
            raise ValueError("Incomplete arguments. Need to specify a factor ID.")
```

workos/mfa.py
```diff
@@ -88,6 +89,59 @@ def enroll_factor(
...
+        if id is None:
+            raise ValueError("Incomplete arguments. Need to specify a factor ID.")
+
+        response = self.request_helper.request(
+            "auth/factors/{id}",
+            method=REQUEST_METHOD_DELETE,
+            token=workos.api_key,
+        )
+
+        return response
```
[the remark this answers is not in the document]
awolfden: Removed the response ✅

workos/mfa.py
```diff
@@ -88,6 +89,59 @@ def enroll_factor(
 
         return WorkOSAuthenticationFactorSms.construct_from_response(response).to_dict()
 
+    def get_factor(
+        self,
+        id=None,
```
blairlunceford, line 94: Can we rename this `authentication_factor_id` to match the other methods? (also applies to the `delete_factor` method)

workos/mfa.py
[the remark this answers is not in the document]
blairlunceford: It looks like the response is still there.

You can use this method as a guid for what to return here: https://github.com/workos/workos-python/blob/main/workos/sso.py#L252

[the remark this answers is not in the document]
awolfden: Updated

workos/mfa.py
awolfden: Updated

blairlunceford (approved):
🎉

PR #153, opened 2022-10-11
Add job title to directory user

blairlunceford:
Add job title to directory user

### DIFF
--- a/workos/resources/directory_sync.py
+++ b/workos/resources/directory_sync.py
@@ -75,6 +75,7 @@ class WorkOSDirectoryUser(WorkOSBaseResource):
         "organization_id",
         "first_name",
         "last_name",
+        "job_title",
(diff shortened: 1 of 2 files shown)


workos/resources/directory_sync.py
```diff
@@ -75,6 +75,7 @@ class WorkOSDirectoryUser(WorkOSBaseResource):
         "organization_id",
         "first_name",
         "last_name",
+        "job_title",
```
maxdeviant, line 78: Does this already correctly account for nullable fields?

blairlunceford: Yes, it looks like this update sets the field to `None` in that case: https://github.com/workos/workos-python/pull/103

maxdeviant (approved):

Issue #144, opened 2022-09-02
ConnectionType enums are not valid in `has_value()`


joetheone:
Hello,

I used to have some code that looked like this:

    authorization_url = workos_client.sso.get_authorization_url(
        provider=ConnectionType.GoogleOAuth,
        redirect_uri=f'{API_HOST}/v1/sso/finish/',
    )

Everything worked well until I updated to from 1.14.0 to 1.15.0.

Now I get the following error: `ValueError: 'provider' must be of type ConnectionType`

This is because `ConnectionType.has_value(ConnectionType.GoogleOAuth)` is False.
Using `ConnectionType.has_value('GoogleOAuth')` works, but it seems like a worse practice than what I had before.

bryanbanda:
Yup, definitely looks like a regression and a worse off practice

maxdeviant:
Thanks for reporting this! This is definitely a regression in the SDK.

It looks like @bryanbanda has a PR up that fixes the issue, so we'll be looking to get that merged soon.

maxdeviant:
This has been addressed in https://github.com/workos/workos/pull/19909 and released in [v1.19.0](https://github.com/workos/workos-python/pull/157).

PR #155, opened 2022-10-17
Correctly parse both string/enum based connection types

bryanbanda:
### Changelog
- Fixes regression introduced by this validation change #137  and handles both string and enum values connection type until the `str`-based one is deprecated. Address issue https://github.com/workos/workos-python/issues/144
- Fixes typo in the pagination order module
- Removes the use of `cls._value2member_map_` in the `ConnectionType` enum which seems to be available only in Python 2.x versions
- Use `raise from` when raising `ValueError` to maintain context and the cause of the error


workos/sso.py
```diff
@@ -213,13 +223,11 @@ def list_connections(
                     "Passing a string value as the 'connection_type' parameter for 'list_connections' is deprecated and will be removed in the next major version. Please pass a 'ConnectionType' instead.",
                     DeprecationWarning,
                 )
-            except KeyError:
-                raise ValueError("'connection_type' must be a member of ConnectionType")
+            except KeyError as key_error:
+                raise ValueError("'connection_type' must be a member of ConnectionType") from key_error
```
maxdeviant, line 227: CI is failing due this this being considered invalid syntax in Python 2.7:

```
./workos/sso.py:227:93: E999 SyntaxError: invalid syntax
                raise ValueError("'connection_type' must be a member of ConnectionType") from key_error
```

We haven't yet dropped supported for Python 2 in the WorkOS SDK, so we'll need to maintain compatibility for the time-being.

maxdeviant:
Thank you for contributing this PR, @bryanbanda!

The changes look good to me, so once CI is passing this should be good to merge.

Aside from one syntactic issue in the Python 2.7 build it looks like you may just need to format the code with Black.

workos/sso.py
bryanbanda: Thanks @maxdeviant, updated! 

Happy to add pre-commit hooks with black etc in a separate PR so that files are correctly formatted on committing changes before CI checks and Tox for running tests across the different supported python versions locally if you'd like.

maxdeviant: > Happy to add pre-commit hooks with black etc in a separate PR

Thanks for the offer, but we're not interested in adding pre-commit hooks to the repo at this time.

maxdeviant (approved):

workos/sso.py
bryanbanda: All good, thanks.

PR #169, opened 2023-02-06
Add auto pagination class methods to list methods

awolfden:
## Description
Add auto pagination class methods to Organizations, Directory Sync, and SSO list methods

## Documentation

```
[ ] Yes
```


workos/resources/organizations.py
```diff
@@ -26,3 +27,39 @@ def to_dict(self):
...
+    @classmethod
+    def construct_from_response(cls, response):
+        return super(WorkOSOrganizationList, cls).construct_from_response(response)
+
+    def to_dict(self):
+        organizations = super(WorkOSOrganizationList, self).to_dict()
+
+        return organizations
+
+    def auto_paging_iter(self):
```
maxdeviant, line 53: If we're going to add auto-pagination we should switch to using our explicit `order` parameter for all the paginating methods so that we can get the new and improved pagination behavior.

This will mean that we can always use `after` to paginate deeper down the list.

blairlunceford (commented):
Would we be able to construct a generic List that the other objects list's (i.e. Connection List, Directory List, etc.) could inherit from? They all seem to have the same methods.

Additionally, would we consider adding a method to get all records for a given list item?

workos/resources/organizations.py
awolfden: Hey @maxdeviant I just refactored this PR to use the order parameter so that we can take advantage of the new behavior and always use the "after" marker to return the next page. 

In order to make sure auto-pagination operates as expected when there is no order parameter submitted, I did make a change to the Python SDK list methods so that it will default to using order=Order.Desc. This allows users to call a list method without defining an order and still be able to auto-paginate. 

If a user is manually paginating and not specifying an order, this could be a breaking change. I'm of the mind that we should push people into the new behavior anyway, so maybe this is a good opportunity to do that. 

Do you have any thoughts about updating the SDK's to use Order.Desc as the default order param if none is submitted to opt everyone in to the new behavior?

awolfden:
This should now be ready for review! 🎉 

The auto pagination method is called `auto_paging_iter` and is a class method on the generic `WorkOSListResource` class. 

This auto-pagination supports both the new Order pagination strategy and the legacy pagination strategy if the order param is left off, so this will not be a breaking change.

tests/test_directory_sync.py
```diff
@@ -283,3 +447,33 @@ def test_primary_email_none(self, mock_user_no_email, mock_request_method):
         me = primary_email.primary_email()
 
         assert me == None
+
+    def test_directories_auto_pagination(self, mock_directories, mock_request_method):
+        mock_request_method("get", mock_directories, 200)
+        directories = self.directory_sync.list_directories(limit=2)
```
blairlunceford, line 453: Should the `limit` be passed here? It seems like it's not being applied, given that the length of `all_directories` is larger than the limit in the check below. If the limit was being honored, I'd expect the length to be 2.

workos/resources/list.py
```diff
@@ -0,0 +1,74 @@
...
+        Returns:
+            dict: A dict representation of the WorkOSListResource
+        """
+        obj_dict = {}
+        for field in self.OBJECT_FIELDS:
+            obj_dict[field] = getattr(self, field, None)
+
+        return obj_dict
+
+    def auto_paging_iter(self):
```
blairlunceford, line 45: Why is there specific logic for each list type? I think the aim here should be a generic list type that could be applied to any object. Can we make this generic - so really just pass through any list options and return all the objects in a given list?

The idea is that when we add new objects, we can just apply a generic list and auto-pagination strategy. So the logic here should not rely on the type of object in the list.

workos/utils/auto_pagination.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def timestamp_compare(timestamp_0, timestamp_1):
+    if timestamp_0 < timestamp_1:
+        return Order.Asc
+    else:
+        return Order.Desc
+
+
+def get_response(type, after, order, directory=None):
```
blairlunceford, line 12: Similar to the comment above, the logic in the list and auto-pagination should be generic, and not reference any specific object type.

tests/test_directory_sync.py
```diff
@@ -43,18 +46,144 @@ def mock_users(self):
                         ],
                     },
                     "id": "directory_usr_id",
-                }
+                    "object": "directory_user",
```
blairlunceford, line 49: Thought on the tests - is there a way in Python to generate a large amount of objects so we can fully test the auto pagination?

blairlunceford (changes requested):

workos/resources/list.py
blairlunceford: Some more context here - the auto pagination method of the list object should just flow through any parameters provided and call the list method for that object, iterating through all the pages, and returning all the resources that apply to the list query.

The only parameters that the auto pagination method needs to adjust are the pagination specific ones that allow it to iterate through the pages of resources returned. But the other parameters should be applied in a generic way by the method.

In this way, the method can be generic, because the pagination parameters are the same across list resources. The pagination should only deal with generic parameters, and just pass through any resource-specific parameters.

tests/test_directory_sync.py
awolfden: The limit being passed here is to ensure that the initial list_directories call does not return the full list. It's not being used in the final assertion. 

That first list_directories call returns the first two values, and then the auto_paging_iter method returns the rest as all_directories. If all_directories matches the mock_directories list it pulled from then the test passes. 

I know there must be a better way to test this, but I do think the limit is being applied successfully. I'll look into programmatically creating a more robust mock list.

workos/resources/list.py
awolfden: Good call. I think this was more of a holdover from how I had initially constructed the method. I refactored it to be more generic at the WorkOSListResource level.

workos/utils/auto_pagination.py
awolfden: Updated `directory` to `parent_resource_id`

tests/test_directory_sync.py
awolfden: I'll look in to this 👀 I think you're right that we can improve on how these tests are structured with a larger mock data set.

tests/test_directory_sync.py
blairlunceford: The limit and other query params should be honored with the auto pagination. The auto pagination should only concern itself with paging through the list, but should flow through any parameters sent.

workos/utils/auto_pagination.py
blairlunceford: Can we remove all references to specific object types in the pagination logic? The idea here is that the list object and auto-pagination wouldn't reference anything specific in the object type - only the parameters needed to page through the list, which are standard across all object types (`after`, `before`, and `order`). It should apply any other query params, but for those specific to the object, they should be generically applied.

The idea behind this is if we add a new resource, or add an additional query parameter to an existing resource, we don't need to update the list object or the pagination methods. We would only need to update these if something changes in the list or pagination as a whole.

For example, in the Stripe Python SDK, the list object is purely generic, so it can be applied to any object type, but doesn't reference any specific object type: https://github.com/stripe/stripe-python/blob/master/stripe/api_resources/list_object.py

workos/resources/list.py
```diff
@@ -0,0 +1,88 @@
...
+        Returns:
+            dict: A dict representation of the WorkOSListResource
+        """
+        obj_dict = {}
+        for field in self.OBJECT_FIELDS:
+            obj_dict[field] = getattr(self, field, None)
+
+        return obj_dict
+
+    def auto_paginate(self):
```
blairlunceford, line 41: This is looking great!

The only question I have is that I think we discussed making this an `auto_paging_iter` function, rather than a function that loads all the resources.

In the case of an `auto_paging_iter` function, it could be used to iterate through whatever list of objects is returned by the list query, so it's more versatile.

It would also only load resources when iterating through, which may help from overloading the API with many requests at once.

WDYT?

blairlunceford: This is the reference method from the Stripe SDK I'm thinking of (obviously not a 1:1 comparison): https://github.com/stripe/stripe-python/blob/master/stripe/api_resources/list_object.py#L80

awolfden: I may have misinterpreted what you were talking about when we were chatting about making it an iterator function, apologies if that's the case! As I understand it an iterator function is one that generates a sequence of values on the fly, allowing you to iterate over a large dataset or process a stream of data one item at a time without having to load the entire dataset into memory.

In the last commit, I implemented it incorrectly since it was still building the full list and returning that list as a generator object, which was not correct as it still stores the full list in memory. With this commit, the generator object is created and each page is returned/yielded to the object as it iterates through each page without needing to store the entire list in memory. The method isn't structured exactly like the one from the stripe SDK, but accomplishes the same thing. 

With the method type passed in as part of self now in the metadata, the function is agnostic of the type of list, so it addresses the versatility goal in that way. 

Let me know if I'm off base or want to chat about it further!

vladmoroz (commented):
Would this need updates to docs?

awolfden:
> Would this need updates to docs?

Hey @vladmoroz 👋🏽 - Yes, when approved we should document the method's availability with the Python SDK as well as update the return from the list methods with Python to include the new metadata dictionary. Would we want to wait until all of the SDK's are updated before making the changes to the docs?

workos/resources/list.py
awolfden: I can refactor it a bit so that auto_paging_iter() can be called directly on the resource like in your example, but it will be a breaking change. Right now for all Python SDK methods, we are returning a dict and not the resource itself. To call the method directly like this, we'll need to adjust the return of the list methods to return the resource and not the resource as a dictionary.

Do we want to have a breaking change and do a major version bump as part of this PR? With the primary_email method addition we had decided against structuring it in this way to avoid a breaking change.

blairlunceford: Good point. I have an idea - what if we don't alter the current list methods on the objects (so keeps them backwards compatible), but just add new list methods that can invoke the `auto_paging_iter` and use the new list object. So this would look like:

```
organizations = Organizations.list
for organization in organizations.auto_paging_iter():
  // do something with the orgs
```

Then we'd keep backwards compatibility, we can set the old list methods as deprecated, and the next time we release a major version, we could potentially remove them.

The only issue might be the naming for connections because they are in the SSO object.

awolfden: I think that'll work nicely! I'll circle back to this likely on Monday. 👍🏽

blairlunceford (approved):
🎉

workos/organizations.py
```diff
@@ -45,24 +47,109 @@ def list_organizations(
...
+            "params": params,
+            "method": Organizations.list_organizations,
+        }
+
+        if "default_limit" in locals():
+            response["metadata"]["params"]["default_limit"] = default_limit
+
+        return response
+
+    def list_organizations_v2(
```
maxdeviant, line 92: How does `list_organizations_v2` differ from the current `list_organizations`?

And what's our strategy for migrating off of the old `list_organizations`?

awolfden: The difference is that the v2 method is returning an instance of the class instead of a dictionary. The thought is that it is a better developer experience to call auto pagination directly on the return variable, ex: list_return.auto_paging_iter() which would not be possible on a dictionary. 

As far as a plan to migrate off the old version, we can support both versions until the next major release, at which point the original version would be removed and we'd only support the new version.

PR #176, opened 2023-05-11
Update semaphore config to specify versions of requests and urllib3

awolfden:
## Description
Update semaphore config to specify versions of requests and urllib3

## Documentation

```
[ ] Yes
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,8 @@ if sys.version_info.major == 3:
         [
             "black==22.3.0",
             "twine==3.1.1",
+            "requests==2.30.0",
+            "urllib3==2.0.2",
         ]
     )
 


AshGodfrey (approved):

setup.py
```diff
@@ -22,6 +22,8 @@
         [
             "black==22.3.0",
             "twine==3.1.1",
+            "requests==2.30.0",
+            "urllib3==2.0.2",
```
amygdalama, line 26: out of curiosity, what's the context for needing this now?

amygdalama (approved):

setup.py
awolfden: Hey @amygdalama thanks for the review! I was running into issues with twine publishing the latest version to PyPI. Based on what I could find while researching the error from the stacktrace, updating these versions should resolve it. 

[Here's a link to the CI run](https://workos.semaphoreci.com/workflows/229dfced-cce0-4cf5-b5be-6fcabdaad28f?pipeline_id=b382ca1a-d99a-4455-ba02-354990df8cdd) where the twine publish failed.

amygdalama: did some digging to validate!

stacktrace (I know you already saw this but just for reference while I take notes here...)
```
Traceback (most recent call last):00:00
  File "/home/semaphore/.pyenv/3.7/bin/twine", line 8, in <module>00:00
    sys.exit(main())00:00
  File "/home/semaphore/.pyenv/3.7/lib/python3.7/site-packages/twine/__main__.py", line 25, in main00:00
    return dispatch(sys.argv[1:])00:00
  File "/home/semaphore/.pyenv/3.7/lib/python3.7/site-packages/twine/cli.py", line 74, in dispatch00:00
    return main(args.args)00:00
  File "/home/semaphore/.pyenv/3.7/lib/python3.7/site-packages/twine/commands/upload.py", line 130, in main00:00
    return upload(upload_settings, args.dists)00:00
  File "/home/semaphore/.pyenv/3.7/lib/python3.7/site-packages/twine/commands/upload.py", line 58, in upload00:00
    repository = upload_settings.create_repository()00:00
  File "/home/semaphore/.pyenv/3.7/lib/python3.7/site-packages/twine/settings.py", line 327, in create_repository00:00
    self.disable_progress_bar,00:00
  File "/home/semaphore/.pyenv/3.7/lib/python3.7/site-packages/twine/repository.py", line 73, in __init__00:00
    self.session.mount(scheme, self._make_adapter_with_retries())00:00
  File "/home/semaphore/.pyenv/3.7/lib/python3.7/site-packages/twine/repository.py", line 84, in _make_adapter_with_retries00:00
    status_forcelist=[500, 501, 502, 503],00:00
TypeError: __init__() got an unexpected keyword argument 'method_whitelist'
```

* twine at version 3.1.1 uses this old `method_whitelist` parameter: https://github.com/pypa/twine/blob/3.1.1/twine/repository.py#L83
* urllib3 renamed this parameter to`allowed_methods` in v2.0.0 https://github.com/urllib3/urllib3/releases/tag/2.0.0
* previously, requests required urllib3 <1.27 but recently (in requests v2.30.0) they added support for urllib3 2.0, which removed this requirement: https://github.com/psf/requests/releases/tag/v2.30.0 (specific change here: https://github.com/psf/requests/commit/2ad18e0e10e7d7ecd5384c378f25ec8821a10a29#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R64)

ok so in summary we didn't have the requests version pinned, but we _did_ have the twine version pinned, and there was a recent requests release that introduced a breaking change.

given the above, I would be surprised if using twine v3.1.1 will work with requests v2.30.0, but maybe I'm missing something! assuming it doesn't work, we can either pin the older versions in all of twine, requests, and urllib3, or pin the newer versions.

I'd probably recommend the newer versions, just so using the old versions doesn't bite us later.

amygdalama: also -- a larger topic -- the above makes me wonder if we want to run tests against both requests <2.30.0 and requests>=2.30.0 if we intend to support them both... I'm not sure if we've thought about this previously? I think other libraries use tox for this: https://tox.wiki/en/latest/

PR #180, opened 2023-05-16
Add method to `get_events` from `Events` service

alisherry:
## Description
Add support to get a list of events from the `Events` service. Introduces a new module for `Events`, and the corresponding endpoint. 

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/events.py
+++ b/workos/events.py
@@ -33,12 +33,12 @@ class Events(WorkOSListResource):
         rangeEnd=None,
     ):
         """Gets a list of Events .
-        Args:
-            events (list): Filter to only return events of particular types.
-            limit (int): Maximum number of records to return.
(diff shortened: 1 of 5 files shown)


workos/events.py
```diff
@@ -0,0 +1,78 @@
...
+    def get_events(
+        self,
+        events=None,
+        limit=None,
+        after=None,
+        rangeStart=None,
+        rangeEnd=None,
+    ):
+        """Gets a list of Events .
+        Args:
```
tribble, line 36: In some places we use `Args` and in other `Kwargs`. Do you know which is correct? =)

tribble (approved):

workos/events.py
alisherry: Ah good callout, I think `Kwargs` would be appropriate here

PR #190, opened 2023-08-04
v1.27.0

pliao28:
## Description
Version bump for fix to add `expires_in` and remove `connections` arg to create passwordless session call.
## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/__about__.py
+++ b/workos/__about__.py
@@ -12,7 +12,7 @@ __package_name__ = "workos"
 
 __package_url__ = "https://github.com/workos-inc/workos-python"
 
-__version__ = "1.26.1"
+__version__ = "1.27.0"
 
 __author__ = "WorkOS"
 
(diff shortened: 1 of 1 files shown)


maxdeviant:
Seems like this should be a minor version bump to v1.27.0, since we're adding functionality (and also removing functionality, which is technically a breaking change, but I just checked and it seems like there is no usage of the `connection` parameter from the Python SDK).

pliao28:
> Seems like this should be a minor version bump to v1.27.0, since we're adding functionality (and also removing functionality, which is technically a breaking change, but I just checked and it seems like there is no usage of the `connection` parameter from the Python SDK).

Yeah I can bump it back up to 1.27.0 then. We don't list the use of this connection param so no one should be using it.

maxdeviant (approved):

PR #209, opened 2023-08-31
Fix Verify Email Code method

jonatascastro12:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/users.py
+++ b/workos/users.py
@@ -23,7 +23,7 @@ USER_AUTHENTICATE_PATH = "users/authenticate"
 USER_PASSWORD_RESET_CHALLENGE_PATH = "users/password_reset_challenge"
 USER_PASSWORD_RESET_PATH = "users/password_reset"
 USER_SEND_VERIFICATION_EMAIL_PATH = "users/{0}/send_verification_email"
-USER_VERIFY_EMAIL_PATH = "users/verify_email"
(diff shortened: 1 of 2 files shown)


jonatascastro12:
Current dependencies on/for this PR:
* users
  * **PR #209** <a href="https://app.graphite.dev/github/pr/workos/workos-python/209" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/workos/workos-python/209?utm_source=stack-comment).

workos/users.py
```diff
@@ -503,14 +503,14 @@ def verify_email(
...
-            USER_VERIFY_EMAIL_PATH,
+            USER_VERIFY_EMAIL_CODE_PATH,
             method=REQUEST_METHOD_POST,
             headers=headers,
             params=payload,
             token=workos.api_key,
         )
 
-        return WorkOSUser.construct_from_response(response).to_dict()
+        return WorkOSUser.construct_from_response(response["user"]).to_dict()
```
ameesha, line 513: just to make sure we want to hard fail if for some reason the "user" key is missing right? (else we'd do .get('user', {})

ameesha (approved):

workos/users.py
jonatascastro12: Yeah... Since this is our API response, I think it's better than having a silent issue that may be hard to debug/understand.

PR #217, opened 2023-11-20
Add auth factor methods

blairlunceford:
## Description

Adds methods to enroll an authentication factor and list the authentication factors for a user.

## Documentation

```
[ ] Yes
```


workos/resources/mfa.py
```diff
@@ -115,3 +115,51 @@ def to_dict(self):
         verification_response_dict = super(WorkOSChallengeVerification, self).to_dict()
 
         return verification_response_dict
+
+
+class WorkOSAuthenticationChallengeAndFactor(WorkOSBaseResource):
```
amygdalama, line 120: i would recommend naming this something like `EnrollAuthFactorResponse` instead (existing example: https://github.com/workos/workos-python/blob/fb0e087584c47f3fdc3ac29e137733b59fe2a15a/workos/resources/password_challenge_response.py#L5 )

or, leaving as two separate classes and building the response in `enroll_auth_factor` (this is what the node SDK does: https://github.com/workos/workos-node/blob/c46747ba60114bed4e0c091a3df7e97ce5a6c894/src/user-management/user-management.ts#L274)

workos/resources/mfa.py
```diff
@@ -115,3 +115,51 @@ def to_dict(self):
...
+
+        authentication_challenge_and_factor.authentication_factor = (
+            WorkOSAuthenticationFactorTotp.construct_from_response(
+                response["authentication_factor"]
+            )
+        )
+
+        return authentication_challenge_and_factor
+
+    def to_dict(self):
```
amygdalama, line 152: do we use this anywhere? seems like we'd only need to go from json -> response object for this case, and not object -> json?

workos/user_management.py
```diff
@@ -540,3 +541,72 @@ def send_magic_auth_code(
...
+            token=workos.api_key,
+        )
+
+        return WorkOSAuthenticationChallengeAndFactor.construct_from_response(
+            response
+        ).to_dict()
+
+    def list_auth_factors(
+        self,
+        user,
```
amygdalama, line 589: do we tend to use params like `user` or `user_id` in the python SDK more? what about within the user management module?

workos/resources/mfa.py
amygdalama: ohh i see we use `to_dict` down here, i missed it https://github.com/workos/workos-python/pull/217/files#diff-87ef5545759351bbeeba22da6f30add44d7c735f751c4807fd059d6ef87931c6R585

wat? lol

amygdalama: ok, i think i see what is happening. we are going from dict -> Response object -> dict. i assume to control the shape of the dict we return from `enroll_auth_factor` ?

amygdalama: it's certainly weird but matches what we currently do in the python SDK so ¯\\_(ツ)_/¯

blairlunceford: Yeah I'm not sure where the pattern derives from, but I just followed what I saw in other resources.

workos/user_management.py
blairlunceford: It seems to be more like `user`, `connection`, `organization`, etc. It definitely seems like doing the `_id` would be better, but that is the convention we're currently using elsewhere:

https://github.com/workos/workos-python/blob/fb0e087584c47f3fdc3ac29e137733b59fe2a15a/workos/user_management.py#L75

https://github.com/workos/workos-python/blob/fb0e087584c47f3fdc3ac29e137733b59fe2a15a/workos/sso.py#L175

https://github.com/workos/workos-python/blob/fb0e087584c47f3fdc3ac29e137733b59fe2a15a/workos/organizations.py#L157

amygdalama: and we use organization_id here... but that's a list endpoint https://github.com/workos/workos-python/blob/fb0e087584c47f3fdc3ac29e137733b59fe2a15a/workos/user_management.py#L93

amygdalama: blegh! lol

workos/resources/mfa.py
blairlunceford: I went with the Node example 👍

workos/user_management.py
blairlunceford: yikes 🙈 I'm ok with starting to be more opinionated and go the `*_id` route

workos/resources/mfa.py
amygdalama: beautiful 🎨

workos/user_management.py
amygdalama: yeah how about within the UM module we go w/ `*_id`?

fwiw we were also inconsistent at the API layer but for UM have made all params `*_id` (and that'll be the plan for new APIs going forward)

blairlunceford: cool cool - I'll make those changes throughout the module now 👍

blairlunceford: I made those changes throughout the UM module, just to remind myself for later changes

amygdalama (approved):
ty!

PR #218, opened 2023-11-20
Add organization membership API methods

blairlunceford:
## Description

Adds methods for organization methods. Removes deprecated methods for adding and removing users from organizations.

## Documentation

```
[ ] Yes
```


tests/test_user_management.py
```diff
@@ -265,21 +292,54 @@ def test_update_user_password(self, mock_user, capture_and_mock_request):
...
-        url, _ = capture_and_mock_request("delete", mock_user, 200)
+        om = self.user_management.get_organization_membership("om_ABCDE")
 
-        user = self.user_management.remove_user_from_organization("user_123", "org_123")
+        assert url[0].endswith("user_management/organization_memberships/om_ABCDE")
+        assert om["id"] == "om_ABCDE"
 
-        assert url[0].endswith("users/user_123/organization/org_123")
-        assert user["id"] == "user_01H7ZGXFP5C6BBQY6Z7277ZCT0"
+    def test_list_organization_memberships_returns_metadata(
```
amygdalama, line 322: what's the purpose of this test?

workos/resources/authentication_response.py
```diff
@@ -1,5 +1,5 @@
 from workos.resources.base import WorkOSBaseResource
-from workos.resources.users import WorkOSUser
+from workos.resources.user_management import WorkOSUser
 
 
 class WorkOSAuthenticationResponse(WorkOSBaseResource):
```
amygdalama, line 5: nice catch on the renaming.

could you move this and the `WorkOSPasswordChallengeResponse` into the `workos.resources.usermanagement` module? (can be in a followup)

tests/test_user_management.py
blairlunceford: I copied this from `list_users` - my understanding is that it's testing that the list method is parsing the data and adding the metadata. But yeah let me look at our tests for list methods and see what might be appropriate here.

workos/user_management.py
```diff
@@ -207,49 +208,125 @@ def update_user_password(self, user_id, password):
...
             token=workos.api_key,
         )
 
-        return WorkOSUser.construct_from_response(response).to_dict()
+        response["metadata"] = {
+            "params": params,
+            "method": UserManagement.list_organization_memberships,
+        }
+
+        if "default_limit" in locals():
```
amygdalama, line 311: lol woof, i know this is copy/pasted and not user-facing so we shouldn't get distracted by this, but this technique could use some work 😆  `default_limit` will always be present in `locals()` since it is set up here 🤦‍♀️  https://github.com/workos/workos-python/pull/218/files#diff-87ef5545759351bbeeba22da6f30add44d7c735f751c4807fd059d6ef87931c6R277

amygdalama: let's ignore for now! 🙃 🙂 🙃 🙂 🙃 🙂 🙃 🙂

tests/test_user_management.py
amygdalama: yeah ideally we'd have a test for the list method that asserts on the user-facing parts

workos/user_management.py
blairlunceford: I'll make a note to look at it in more detail once we get the UM API out 👍

workos/resources/authentication_response.py
blairlunceford: good catch - I moved those 👍

workos/user_management.py
amygdalama: 🙌 happy to look at it with you when the time comes!

amygdalama (approved):
thank you!!

PR #222, opened 2023-11-30
Finalize create and update users API

blairlunceford:
## Description

Finalizes the create and update users API methods

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/user_management.py
+++ b/workos/user_management.py
@@ -22,7 +22,6 @@ USER_PATH = "user_management/users"
 USER_DETAIL_PATH = "user_management/users/{0}"
 ORGANIZATION_MEMBERSHIP_PATH = "user_management/organization_memberships"
 ORGANIZATION_MEMBERSHIP_DETAIL_PATH = "user_management/organization_memberships/{0}"
-USER_PASSWORD_PATH = "users/{0}/password"
 USER_AUTHENTICATE_PATH = "user_management/authenticate"
 USER_PASSWORD_RESET_CHALLENGE_PATH = "users/password_reset_challenge"
 USER_PASSWORD_RESET_PATH = "users/password_reset"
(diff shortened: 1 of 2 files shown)


workos/user_management.py
```diff
@@ -160,27 +133,44 @@ def list_users(
...
             payload (dict) - The User attributes to be updated
-                user[first_name] (string) - The user's first name.
-                user[last_name] (string) - The user's last name.
-                user[email_verified] (bool) - Whether the user's email address was previously verified.
+                payload[first_name] (str) - The user's first name. (Optional)
+                payload[last_name] (str) - The user's last name. (Optional)
+                payload[email_verified] (bool) - Whether the user's email address was previously verified. (Optional)
+                payload[password] (str) - The password to set for the user. (Optional)
+                payload[password_hash] (str) - The hashed password to set for the user, used when migrating from another user store. Mutually exclusive with password. (Optional)
+                payload[password_hash_type] (str) - The algorithm originally used to hash the password, used when providing a password_hash. Valid values are bcrypt. (Optional)
```
oliverzheng, line 173:
```suggestion
                payload[password_hash_type] (str) - The algorithm originally used to hash the password, used when providing a password_hash. Only valid value is 'bcrypt'. (Optional)
```

oliverzheng (approved):

workos/user_management.py
```diff
@@ -160,27 +133,44 @@ def list_users(
...
-        self.request_helper.request(
-            USER_DETAIL_PATH.format(user_id),
-            method=REQUEST_METHOD_DELETE,
+        headers = {}
+
+        response = self.request_helper.request(
+            USER_PATH,
+            method=REQUEST_METHOD_POST,
+            params=user,
+            headers=headers,
```
amygdalama, line 156: ooc what was this change (adding empty headers) for? or was it copy/paste?

blairlunceford: This was me moving methods around (so they are in a different order), and the diff just not dealing with that well 🙃

amygdalama: oooooh yeah i see my bad 😆

PR #224, opened 2023-12-05
Add UM `get_authorization_url` method

blairlunceford:
## Description

Adds the UM specific `get_authorization_url` method.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/user_management.py
+++ b/workos/user_management.py
@@ -10,6 +10,7 @@ from workos.resources.user_management import (
     WorkOSUser,
 )
 from workos.utils.pagination_order import Order
+from workos.utils.provider_types import ProviderType
 from workos.utils.request import (
     RequestHelper,
     RESPONSE_TYPE_CODE,
@@ -340,7 +341,7 @@ class UserManagement(WorkOSListResource):
                 The value of this parameter should be a WorkOS Connection ID. (Optional)
             organization_id (str) - The organization_id connection selector is used to initiate SSO for an Organization.
                 The value of this parameter should be a WorkOS Organization ID. (Optional)
-            provider (str) - The provider connection selector is used to initiate SSO using an OAuth-compatible provider.
+            provider (ProviderType) - The provider connection selector is used to initiate SSO using an OAuth-compatible provider.
                 Currently, the supported values for provider are 'authkit', 'GoogleOAuth' and 'MicrosoftOAuth'. (Optional)
             domain_hint (str) - Can be used to pre-fill the domain field when initiating authentication with Microsoft OAuth,
                 or with a GoogleSAML connection type. (Optional)
@@ -369,7 +370,10 @@ class UserManagement(WorkOSListResource):
         if organization_id is not None:
             params["organization_id"] = organization_id
         if provider is not None:
-            params["provider"] = provider
+            if not isinstance(provider, ProviderType):
+                raise ValueError("'provider' must be of type ProviderType")
+
+            params["provider"] = provider.value
         if domain_hint is not None:
             params["domain_hint"] = domain_hint
         if login_hint is not None:
--- a/workos/user_management.py
+++ b/workos/user_management.py
@@ -10,7 +10,7 @@ from workos.resources.user_management import (
     WorkOSUser,
 )
 from workos.utils.pagination_order import Order
-from workos.utils.provider_types import ProviderType
+from workos.utils.um_provider_types import UserManagementProviderType
 from workos.utils.request import (
     RequestHelper,
     RESPONSE_TYPE_CODE,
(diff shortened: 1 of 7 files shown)


workos/user_management.py
```diff
@@ -316,6 +319,72 @@ def delete_organization_membership(self, organization_membership_id):
...
+            raise ValueError(
+                "Incomplete arguments. Need to specify either a 'connection_id', 'organization_id', or 'provider_id'"
+            )
+
+        if connection_id is not None:
+            params["connection_id"] = connection_id
+        if organization_id is not None:
+            params["organization_id"] = organization_id
+        if provider is not None:
+            params["provider"] = provider
```
oliverzheng, line 372: We can do the same is-valid-provider check that this does: https://github.com/workos/workos-python/blob/91a24b315767ba86a19b373aca17ec1fc5608714/workos/sso.py#L87-L101

Since it's a parameter that's used as input to the API, any new additions would not result in SDK-breaking changes; existing SDKs will simply not be able to use new features, which is fine. (It's when an enum value that comes out of the API as a response that becomes problematic if checked at runtime.)

workos/user_management.py
```diff
@@ -316,6 +319,72 @@ def delete_organization_membership(self, organization_membership_id):
...
+        if domain_hint is not None:
+            params["domain_hint"] = domain_hint
+        if login_hint is not None:
+            params["login_hint"] = login_hint
+        if state is not None:
+            params["state"] = state
+
```
oliverzheng, line 379: Should we throw the same error if redirect_uri is not provided? I think the user management API still requires `redirect_uri`. https://github.com/workos/workos-python/blob/91a24b315767ba86a19b373aca17ec1fc5608714/workos/sso.py#L120-L121

workos/user_management.py
blairlunceford: Just to confirm here - this is a check against the ConnectionType enum, defined here: https://github.com/workos/workos-python/blob/main/workos/utils/connection_types.py.

I believe unless the input is a defined enum value, we will throw an error: https://github.com/workos/workos-python/blob/d2e71041a42f152c1da8e9779499f75fa3e3b3e1/tests/test_sso.py#L207

The supported values for provider are `GoogleOAuth`, `MicrosoftOAuth`, and `authkit`. I can add `authkit` to the enum, but I don't think any new values would be supported unless added to the enum.

I actually question whether we should have this check for the SSO method, given that the ConnectionType spans more values than are actually supported in the `provider`, so it would be better to make a different enum or just go with a string.

For reference, I used the Node SDK as reference, and we just have it as a string type there: https://github.com/workos/workos-node/blob/abb683ccf55c79b7bb48639ed591fd4241f12f1e/src/user-management/interfaces/authorization-url-options.interface.ts#L7

workos/user_management.py
blairlunceford: I have it as a required param, so a TypeError will be thrown if it's not provided. I added a test for this case 👍

workos/user_management.py
oliverzheng: Ah I see. The API used to (and I guess still does) accept connection types for `provider` despite not publicly documenting that it does, which makes sense. I don't think we are cutting a new major version for the Python SDK so it doesn't make sense to change the SSO API.

It would make sense if we can have something like a `ProviderType` so that we can check against valid values for the User Management authorize API. And since `authkit` is not actually a Connection type, it wouldn't make sense to add it as a value there.

When we cut a major version for this SDK, we can change the SSO authorize url behavior as well so not all connection types can be passed into `provider`.

> For reference, I used the Node SDK as reference, and we just have it as a string type there.

It'd be nice if we had types there too. Arguably, it'd be a breaking SDK change if a provider value that was invalid suddenly became valid. I.e. if we add a new provider type in the API in the future, the SDK would not need a change, but previous failing requests would now succeed. Realistically, no one would care, but having enums for inputs into the API is always nice.

blairlunceford: Sounds good - the plan is to cut a major version, so I'll go ahead and create a ProviderType to use as an input for both SSO and UserManagement methods 👍

oliverzheng (commented):
Looks good! Just one thing about the varying type by entry point.

kvnkho:
Hi, I'm trying to use AuthKit and I think I need this PR to be able to with the WorkOS Python Client. I know I could just clone master, but I was wondering if you had an idea when this would be released to PyPI?

Thank you for your work!

kvnkho:
Awesome, thank you for the info! The code is very readable so I was able to hack it. Still would prefer using the official one though. :D

blairlunceford:
@kvnkho the latest major release (2.0.0) is on PyPi and includes the User Management API. Let us know if you have any feedback as you use this new feature. Thanks!

kvnkho:
Thank you for responding. Will try it out!

kvnkho:
Works great! Python docs just need tweaking as it says:

```python
authorization_url = client.sso.get_authorization_url(
    connection="conn_01E4ZCR3C56J083X43JQXF3JK5",
    redirect_uri="https://your-app.com/callback",
    state="dj1kUXc0dzlXZ1hjUQ==",
 )
```

But AuthKit is actually under `client.user_management.get_authorization_url` so I guess it can be a bit confusing having two `get_authorization_url` method where one seems to be the superset of the other (totally not sure). Worked for me when I did:

```python
    authorization_url = client.user_management.get_authorization_url(
        provider=UserManagementProviderType.AuthKit,
        redirect_uri="http://localhost:5050/auth/authenticate",
        state=None,
    )
```

PR #234, opened 2024-03-05
chore: default list methods to descending order

faroceann:
## Description
- Default list methods to descending order. 
- Remove tests on python v2 

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/directory_sync.py
+++ b/workos/directory_sync.py
@@ -84,7 +84,9 @@ class DirectorySync(WorkOSListResource):
                 params["order"] = order
             else:
                 raise ValueError("Parameter order must be of enum type Order")
-
+        else: 
+            params["order"] = "desc"
+            
         response = self.request_helper.request(
             "directory_users",
             method=REQUEST_METHOD_GET,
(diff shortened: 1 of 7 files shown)


workos/user_management.py
```diff
@@ -287,6 +289,8 @@ def list_organization_memberships(
                 params["order"] = order
             else:
                 raise ValueError("Parameter order must be of enum type Order")
+        else:
```
PavanKulkarni, line 292: can just define `order` as default when params is initialized.

workos/directory_sync.py
```diff
@@ -84,6 +84,8 @@ def list_users(
                 params["order"] = order
             else:
                 raise ValueError("Parameter order must be of enum type Order")
+        else:
+            params["order"] = "desc"
```
alisherry, line 88: above in `params`, 

can we default to `desc? 

e.g 

```  
 params = {
            "limit": limit,
            "before": before,
            "after": after,
            "order": order or "desc",
        }
```

alisherry (approved):

PavanKulkarni (approved):

PR #247, opened 2024-04-03
Python SDK: Add 25 second default timeout to requests

awolfden:
## Description
Python SDK: Add 25 second default timeout to requests

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/utils/request.py
+++ b/workos/utils/request.py
@@ -25,8 +25,6 @@ REQUEST_METHOD_GET = "get"
 REQUEST_METHOD_POST = "post"
 REQUEST_METHOD_PUT = "put"
 
-REQUEST_DEFAULT_TIMEOUT = workos.request_timeout
-
 
 class RequestHelper(object):
     def __init__(self):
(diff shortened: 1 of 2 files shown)


workos/utils/request.py
```diff
@@ -95,11 +100,11 @@ def request(
                 raise AuthenticationException(response)
             elif status_code == 403:
                 raise AuthorizationException(response)
-            elif status_code == 404:
```
PaulAsjes, line 98: Why remove this?

PaulAsjes (changes requested):

workos/utils/request.py
awolfden: Definitely didn't mean to remove this, fat finger error. Will update.

awolfden: @PaulAsjes updated

workos/utils/request.py
```diff
@@ -99,7 +105,9 @@ def request(
                 raise NotFoundException(response)
             error = response_json.get("error")
             error_description = response_json.get("error_description")
-            raise BadRequestException(response)
+            raise BadRequestException(
+                response, error=error, error_description=error_description
```
PaulAsjes, line 109: More of a nit, but I wonder why we weren't returning these earlier...

PaulAsjes (approved):

Issue #252, opened 2024-04-10
missing string formatting 


yehuda-margolis:
https://github.com/workos/workos-python/blob/a145e1200ff200317dd292177cbfc5e40cb0cc64/workos/mfa.py#L138

The authentication_factor_id argument is sent as is (string) instead of the argument's actual value

amadeo-workos:
https://github.com/workos/workos-python/pull/254

amadeo-workos:
This has been fixed, and included in the v4.3.1 release: https://github.com/workos/workos-python/releases/tag/v4.3.1

Closing this issue now.

PR #254, opened 2024-04-10
Fix string interpolation when building MFA URLs

amadeo-workos:
## Description
This PR aims at fixing the string interpolation when building MFA URLs.

This is to address https://github.com/workos/workos-python/issues/252


jonatascastro12 (approved):
LGTM

PR #256, opened 2024-04-11
Explicit coverage over creation of parameterized URLs

amadeo-workos:
## Description
This is to ensure we have the right helpers (and test coverage) to build URLs in the right way, and not experience code defects as the ones reported recently.

### DIFF
--- a/workos/mfa.py
+++ b/workos/mfa.py
@@ -106,7 +106,10 @@ class Mfa(object):
             raise ValueError("Incomplete arguments. Need to specify a factor ID")
 
         response = self.request_helper.request(
-            self.request_helper.build_parameterized_url("auth/factors/{authentication_factor_id}", authentication_factor_id=authentication_factor_id),
+            self.request_helper.build_parameterized_url(
+                "auth/factors/{authentication_factor_id}",
+                authentication_factor_id=authentication_factor_id,
+            ),
             method=REQUEST_METHOD_GET,
             token=workos.api_key,
         )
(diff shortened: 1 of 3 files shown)


amadeo-workos:
CC @jonatascastro12 @PaulAsjes @southgate

workos/mfa.py
```diff
@@ -106,8 +106,9 @@ def get_factor(
             raise ValueError("Incomplete arguments. Need to specify a factor ID")
 
         response = self.request_helper.request(
-            "auth/factors/{authentication_factor_id}".format(
```
mthadley, line 109: What's the status of [formatted string literals](https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings) in Python and our SDK Python version support? [Looks like it was added in 3.6](https://peps.python.org/pep-0498/) and our CI tests only go back to 3.8.

If we were to write `f"auth/factors/{authentication_factor_id}"` is there still a need for this method? Though I have one idea why we might want it (will leave a separate comment).

workos/utils/request.py
```diff
@@ -45,6 +45,9 @@ def set_base_api_url(self, base_api_url):
     def generate_api_url(self, path):
         return self.base_api_url.format(path)
 
+    def build_parameterized_url(self, url, **params):
+        return url.format(**params)
```
mthadley, line 49: While I wonder whether we can just use formatted string literals, one reason to prefer a helper like this would be to do automatic URL component escaping:

```py
def build_parameterized_url(self, url, **params):
    escaped_params = {k: urllib.parse.quote(str(v)) for k, v in params.items()}
    return url.format(**escaped_params)
```

amadeo-workos: @mthadley The initial idea that triggered the addition of this function was so we could test these interpolations. A few days ago a customer reported a bug caused by the interpolation silently failing.

Now that you bring up the escaping, that's an additional support we could have with this method. Thanks for bringing this up! I committed the suggestion.

mthadley (approved):

PR #264, opened 2024-05-10
Add organization membership deactivate and reactivate API methods

mattgd:
## Description
Add organization membership deactivate and reactivate API methods.

## Documentation

```
[x] Yes
```


workos/user_management.py
```diff
@@ -300,6 +307,7 @@ def list_organization_memberships(
         Kwargs:
             user_id (str): Filter Organization Memberships by user. (Optional)
             organization_id (str): Filter Organization Memberships by organization. (Optional)
+            statuses (list): Filter Organization Memberships by status. (Optional)
```
PavanKulkarni, line 310: do we generally use this naming pattern for list params?

mattgd: Do mean the pluralized parameter name? Seems like we do in the Python SDK:
https://github.com/workos/workos-python/blob/main/workos/organizations.py#L41
https://github.com/workos/workos-python/blob/main/workos/events.py#L38

PavanKulkarni: no strong opinion then but I feel `statuses` seems a bit weird. Maybe `status_list`

mattgd: Down to change it. That being said `statuses` is how it's defined in the API/other SDKs. Maybe it's better to be consistent, thoughts?

PaulAsjes: I think we should be consistent and stick with `statuses`.

PaulAsjes (approved):

PR #275, opened 2024-06-11
Add PKCE parameters to UM auth flow

awolfden:
## Description
Add PKCE parameters to UM auth flow

## Documentation

```
[X] Yes
```

### DIFF
--- a/workos/user_management.py
+++ b/workos/user_management.py
@@ -424,7 +424,6 @@ class UserManagement(WorkOSListResource):
         login_hint=None,
         state=None,
         code_challenge=None,
-        code_challenge_method="S256",
     ):
         """Generate an OAuth 2.0 authorization URL.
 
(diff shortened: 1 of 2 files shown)


workos/user_management.py
```diff
@@ -476,6 +481,9 @@ def get_authorization_url(
             params["login_hint"] = login_hint
         if state is not None:
             params["state"] = state
+        if code_challenge is not None:
```
ameesha, line 484: is an empty string valid? if not you can switch this to `if code_challenge:`

workos/user_management.py
```diff
@@ -444,6 +446,9 @@ def get_authorization_url(
                 OktaSAML, and AzureSAML connection types. (Optional)
             state (str) - An encoded string passed to WorkOS that'd be preserved through the authentication workflow, passed
                 back as a query parameter. (Optional)
+            code_challenge (str) - Code challenge is derived from the code verifier used for the PKCE flow. (Optional)
+            code_challenge_method ("S256") - The only valid PKCE code challenge method is "S256".
```
ameesha, line 450: if there's only 1 option for this field, why pass it in at all? let's just set it only when code_challenge is passed in

workos/user_management.py
awolfden: Updated ✅

workos/user_management.py
awolfden: Updated ✅

workos/user_management.py
```diff
@@ -476,6 +478,9 @@ def get_authorization_url(
             params["login_hint"] = login_hint
         if state is not None:
             params["state"] = state
+        if code_challenge:
+            params["code_challenge"] = code_challenge
+            params["code_challenge_method"] = "S256"
```
ameesha, line 483: quick follow up - since this is a static value, does the client need to pass it in at all?

ameesha (approved):

workos/user_management.py
awolfden: It's listed in the API reference as a param for curl, so I included it here

PR #281, opened 2024-06-27
update list_users function documentation on UserManagement class

AbundantSalmon:
## Description

Had issues using the returned value from `list_users` on the `UserManagement` class, returned values was not a `dict` as described in the function docstring. 

Was going to modify the `list_users` return value in this PR, but then realised with the tests that a documented use case of `auto_paging_iter()` was is expected, as such the documentation should be updated to minimise developer confusion.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/user_management.py
+++ b/workos/user_management.py
@@ -109,7 +109,7 @@ class UserManagement(WorkOSListResource):
             order (Order): Sort records in either ascending or descending order by created_at timestamp: "asc" or "desc" (Optional)
 
         Returns:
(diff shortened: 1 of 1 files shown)


PaulAsjes:
Hey there, thanks for the PR! 

This is definitely an issue but I don't think we should solve it by updating the code comment. I think we need to fix this for good by making sure that what the SDK does matches the docs. We'll fix this by making sure all list functions return a dict.

PR #282, opened 2024-07-02
Add typing for Directory Sync events

ameesha:
## Documentation

```
[ ] Yes
```


workos/event_objects/directory.py
```diff
@@ -0,0 +1,71 @@
...
+        self.state: DirectoryState = DirectoryState.ACTIVE
+        self.domains: List[OrganizationDomain] = []
+        for domain in attributes["domains"]:
+            self.domains.push(OrganizationDomain(domain))
+        self.organization_id: str = attributes["organization_id"]
+        self.created_at: str = attributes["created_at"]
+        self.updated_at: str = attributes["updated_at"]
+        self.external_key: str = attributes["external_key"]
+        # always 'directory' for this event
+        self.object: str = attributes["object"]
```
ameesha, line 55: can also just set this statically (like i do for self.event). uncertain which is better (it's the same both ways though)

workos/event_objects/directory_user.py
```diff
@@ -0,0 +1,77 @@
...
+
+class DirectoryUserRole:
+    def __init__(self, attributes: JsonDict) -> None:
+        self.slug: str = attributes["slug"]
+
+
+class DirectoryUserEmail:
+    def __init__(self, attributes: JsonDict) -> None:
+        self.type: Optional[str] = attributes.get("type")
+        self.value: Optional[str] = attributes.get("value")
```
mattgd, line 14: Is value ever null/None based on our API schema?

workos/event_objects/directory_user.py
```diff
@@ -0,0 +1,77 @@
...
+
+class DirectoryUserState(Enum):
+    ACTIVE = "active"
+    INACTIVE = "inactive"
+
+
+class DirectoryUserEvent:
+    def __init__(self, attributes) -> None:
+        self.id: str = attributes["id"]
+        self.name: str = attributes["name"]
```
mattgd, line 26: What's name here? I don't see it in our docs.

workos/events.py
```diff
@@ -24,6 +68,45 @@ def request_helper(self):
...
+        self,
+        limit: Optional[int] = None,
+        organization_id: Optional[str] = None,
+        after: Optional[str] = None,
+        range_start: Optional[str] = None,
+        range_end: Optional[str] = None,
+    ) -> ListDsyncEventsResponse:
+        if limit is None:
+            limit = RESPONSE_LIMIT
+            default_limit = True
```
mattgd, line 81: Is `default_limit` used?

workos/event_objects/directory.py
```diff
@@ -0,0 +1,74 @@
...
+    PERSONIO = "personio"
+    PINGFEDERATE_SCIM_v2 = "pingfederate scim v2.0"
+    RIPPLING_SCIM_v2 = "rippling v2.0"
+    SFTP = "sftp"
+    SFTP_WORKDAY = "sftp workday"
+    WORKDAY = "workday"
+
+
+class DirectoryState(Enum):
+    ACTIVE = "active"
```
PavanKulkarni, line 30: we never show `validating` right?

workos/event_objects/directory.py
```diff
@@ -0,0 +1,74 @@
...
+    JUMPCLOUD_SCIM_v2 = "jump cloud scim v2.0"
+    OKTA_SCIM_v2 = "okta scim v2.0"
+    ONELOGIN_SCIM_v2 = "onelogin scim v2.0"
+    PEOPLE_HR = "people hr"
+    PERSONIO = "personio"
+    PINGFEDERATE_SCIM_v2 = "pingfederate scim v2.0"
+    RIPPLING_SCIM_v2 = "rippling v2.0"
+    SFTP = "sftp"
+    SFTP_WORKDAY = "sftp workday"
+    WORKDAY = "workday"
```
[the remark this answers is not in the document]
PavanKulkarni: yeah S3, Rippling and Gusto are completely deprecated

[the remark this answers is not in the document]
mattgd: Got it, so we can remove Gusto from this list also.

workos/event_objects/directory.py
mattgd: I don't believe there's ever an directory event emitted with `status = validating` (or any status besides active, deleting).

workos/event_objects/directory_user.py
ameesha: Yeah, I left a comment about this in the docs PR as well.

workos/event_objects/directory_user.py
ameesha: Ooh might be a copy/paste error when I was cleaning up. Good catch

workos/events.py
ameesha: It can get updated (look at other function), I removed it since it seemed unnecessary but didn't get to testing it to confirm

workos/event_objects/directory.py
ameesha: Correct

workos/event_objects/directory.py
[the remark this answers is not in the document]
tribble: According to mode, these are the directory types that actually exist in produiction:
CleanShot 2024-07-02 at 09 36 56

We don't actually have any Personio directories, which is no big deal.

But we do have Rippling directories (non-SCIM) in the DB. That's not reflected here.

We don't actually have any "untyped" directories in prod anymore, so we don't need to worry about this `pending` state: https://github.com/workos/workos/blob/1b4c7a6a83b49ec3a605ff4f2f5dd17848e6085d/packages/api/src/directories/directories.controller.ts#L63

We should have a KTLO item to clean that up.

workos/event_objects/directory.py
tribble: I don't see where we exclude validating directories: https://github.com/workos/workos/blob/1b4c7a6a83b49ec3a605ff4f2f5dd17848e6085d/packages/api-services/src/directories/directories.service.ts#L322

workos/event_objects/directory.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class DirectoryDeletedEvent:
+    event_name = "dsync.deleted"
+
+    def __init__(self, attributes: JsonDict) -> None:
+        self.event: str = attributes["event"]
+        self.id: str = attributes["id"]
+        self.created_at = attributes["created_at"]
+        self.data: DirectoryEvent = DirectoryEvent(attributes["data"])
```
tribble, line 74: We can't use this same type here. The directory deleted event is not the same shape as the directory activated event: https://workos.com/docs/events/directory-sync

It uses the directory type, not the legacy directory type with extra fields: https://github.com/workos/workos/blob/e95130bf78e01b202c7e7417adf36135cfc5f033/packages/api-json-models/src/events/directory-deleted.ts#L9

workos/event_objects/directory.py
tribble: Should the type here [be `LiteralString`](https://docs.python.org/3/library/typing.html#typing.LiteralString)?

tribble: Still learning python typing, but maybe this should be `Literal["directory"]`?

workos/event_objects/directory_group.py
```diff
@@ -0,0 +1,46 @@
...
+class DirectoryGroupEvent:
+    def __init__(self, attributes) -> None:
+        self.id: str = attributes["id"]
+        self.name: str = attributes["name"]
+        self.idp_id: str = attributes["idp_id"]
+        self.directory_id: str = attributes["directory_id"]
+        self.organization_id: str = attributes["organization_id"]
+        self.created_at: str = attributes["created_at"]
+        self.updated_at: str = attributes["updated_at"]
+        self.raw_attributes: JsonDict = attributes.get("raw_attributes", {})
```
tribble, line 13: @PavanKulkarni didn't we stop emitting `raw_attributes` for directory group events?

workos/event_objects/directory.py
mattgd: Via the directory API we definitely return other directory states, but the type is specifically event-scoped where we'd only have active or deleting directories. If we want to enumerate all states, we could extract this enum from the event context.

workos/event_objects/directory_user.py
```diff
@@ -0,0 +1,77 @@
...
+        self.emails: List[DirectoryUserEmail] = []
+        for email in attributes.get("emails"):
+            self.emails.push(DirectoryUserEmail(email))
+        self.username: Optional[str] = attributes.get("username")
+        self.state: DirectoryUserState = DirectoryUserState(attributes["state"])
+        self.custom_attributes: JsonDict = attributes.get("custom_attributes", {})
+        self.raw_attributes: JsonDict = attributes.get("raw_attributes", {})
+        self.created_at: str = attributes["created_at"]
+        self.updated_at: str = attributes["updated_at"]
+        self.role: Optional[DirectoryUserRole] = (
```
tribble, line 42: Is this actually optional anymore?

workos/event_objects/directory.py
mattgd: Yeah, that'd be better. Updating.

workos/event_objects/directory.py
tribble: So, is the idea that we are *only* typing the events and not the dsync APIs?

workos/event_objects/directory.py
tribble: Per the [events JSON models](https://github.com/workos/workos/tree/e95130bf78e01b202c7e7417adf36135cfc5f033/packages/api-json-models/src/events), it is accurate.

As for "here to stay", I think it contains fields that are no longer necessary. But I'm dubious about the dsync team prioritizing an effort to remove the unnecessary fields. So, likely here to stay for a while.

workos/event_objects/directory_user.py
mattgd: I didn't think so, but it's still optional in the API event schema, so will keep as-is until the API is updated.

workos/event_objects/directory.py
mattgd: Ideally we should type all the things, but I think that's worth a separate PR.

workos/event_objects/directory_group.py
PavanKulkarni: Let me confirm this in LD.

PavanKulkarni: yes only 3 teams still depend on the [flag](https://app.launchdarkly.com/projects/default/flags/exclude-group-raw-attributes-outgoing/targeting?env=development&env=test&env=production&selected-env=production). That being said we still send `raw_attributes` IIRC, it's just empty.

workos/event_objects/directory.py
tribble: Sure. But I'm don't think we should release the SDK until both the Dsync events and API resources are typed.

workos/event_objects/directory_user.py
tribble: It's the API resource that led me to ask this question. The role here always exists when returning a directory user: https://github.com/workos/workos/blob/e95130bf78e01b202c7e7417adf36135cfc5f033/packages/api/src/directory-users/directory-users.controller.ts#L340

workos/event_objects/directory.py
mattgd: Yep, agreed.

mattgd: Ended up adding some light typing for the APIs using existing objects. The list endpoints could use some better typing, but would require a size-able rework of the `auto_paging_iter` to do so.

workos/directory_sync.py
```diff
@@ -439,19 +439,19 @@ def list_directories(
...
-        after=None,
-        organization=None,
-        order=None,
+        domain: str = None,
+        search: str = None,
+        limit: int = None,
+        before: str = None,
+        after: str = None,
+        organization: str = None,
+        order: Order = None,
```
tribble, line 448: I'm new to python type hints, but is this correct? Coming from a TS world something like `foo(domain: string = null)` is not valid.

tribble: I'm still learning whether that's OK in Python. It's OK to set a type hint without initializing a variable.


... ah, OK, looks like it's OK to initialize with `None`

mattgd: We could definitely make this better with `Optional`, which I'll do.

tribble: `Optional` does sound more correct here. I think. Says the person that doesn't actually write anything in Python :P

workos/event_objects/directory.py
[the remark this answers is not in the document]
ameesha: This was for the events list so we wouldn't have events with these types anymore.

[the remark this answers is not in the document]
ameesha: Nvm catching up to see we're doing API as well now lol

workos/event_objects/directory_user.py
ameesha: It's optional! We don't return it on non-scim/Google users

ameesha: https://github.com/workos/workos/blob/main/packages/api-services/src/directory-users/utils/serialize-directory-user.ts#L79

workos/events.py
```diff
@@ -78,4 +121,11 @@ def list_events(
             "method": Events.list_events,
         }
 
+        data = []
+        for list_data in response["data"]:
+            if list_data["event"] in EVENT_TO_OBJECT.keys():
```
ameesha, line 126: if we add a new event or update an event type, until the SDK is explicitly updated we'll just ignore it

workos/event_objects/directory_user.py
tribble: @ameesha but in the code I referenced `role` *always* exists because `getRoleFromUserRoleId` returns `Promise<Role>`.

I  see that the serializer accepts `role?`, but in this code path I don't see how it can ever actually happen.

ameesha: yeah the role is always grabbed but in the serializer (line i sent), we only use it for scim/google dirs

ameesha: nothing cares about whether the role is relevant in the directory type or not but the serializer. it's the only* thing that checks that logic (cause we create and set roles across the board). 

* there are other checks for directory type so we can end recalculations early but if we removed it, nothing would really change

PR #284, opened 2024-07-12
Refactor and streamline organization and set base for SDK typing

jonatascastro12:
- Create new `WorkOsListResource` to fix `.auto_pagination_iter()` for list methods
- Add `pydandic` for setting new resource classes for `Organization` and other entities (come later)
- Add `types-requests` as mypy complained for typing `requests` lib
- Add `mypy` (initially scoping to `workos/organizations.py`)
- Refactored `workos/organizations.py` - methods return `pydandic` models
  - removed `.list_organizations_v2()` method since they were meant to work with `.auto_pagination_iter()`, but it wasn't functional
  - create/update methods are consistent with using args
  - typed all methods arguments
- Removed deprecated fields on organizations


workos/organizations.py
```diff
@@ -29,77 +34,12 @@ def request_helper(self):
...
-    def list_organizations_v2(
-        self,
-        domains=None,
-        limit=None,
-        before=None,
-        after=None,
-        order=None,
-    ):
+        domains: Optional[List[str]] = None,
+        limit: Optional[int] = None,
```
tribble, line 38: If `limit` has a default value of `RESPONSE_LIMIT` below, any reason not to make that the default value here?
```suggestion
        limit: Optional[int] = RESPONSE_LIMIT,
```

tribble: ... and remove the `default_limit = True` below. It doesn't seem to do anything.

tests/conftest.py
```diff
@@ -45,7 +47,7 @@ def inner(method, response_dict, status_code, headers=None):
         def mock(*args, **kwargs):
             return MockResponse(response_dict, status_code, headers=headers)
 
-        monkeypatch.setattr(requests, method, mock)
+        monkeypatch.setattr(requests, "request", mock)
```
tribble, line 50: I changed how we call `requests`, rather than using `get_attr` to retrieve the function by name, always wall `request.request` and pass in the method

tests/conftest.py
```diff
@@ -73,8 +75,46 @@ def capture_and_mock(*args, **kwargs):
...
-        monkeypatch.setattr(requests, method, capture_and_mock)
+        monkeypatch.setattr(requests, "request", capture_and_mock)
 
         return (request_args, request_kwargs)
 
     return inner
+
+
+@pytest.fixture
+def mock_pagination_request(monkeypatch):
```
tribble, line 86: I don't like how complex this test function is, but we need to roughly emulate the behavior of a pagination endpoint and return different pages of data in successive calls.

tests/test_directory_sync.py
```diff
@@ -18,110 +22,19 @@ def mock_users(self):
         return {
             "data": user_list,
             "list_metadata": {"before": None, "after": None},
-            "metadata": {
```
tribble, line 21: All of this no longer used/needed

workos/organizations.py
```diff
@@ -47,114 +50,30 @@ def list_organizations(
...
-            limit = RESPONSE_LIMIT
-            default_limit = True
 
         params = {
             "domains": domains,
             "limit": limit,
             "before": before,
             "after": after,
-            "order": order or "desc",
+            "order": order,
```
mattgd, line 59: With `order` previously being passed as `None`, this doesn't change the default ordering at all, right?

workos/resources/directory_sync.py
```diff
@@ -1,102 +1,115 @@
...
-
-class WorkOSDirectoryGroup(WorkOSBaseResource):
+    id: str
+    object: Literal["directory"]
+    domain: Optional[str] = None
+    name: str
+    organization_id: str
+    # Toying around with the differences in type hinting or deserialization for enums vs literals. In the end pretty equivalent,
+    # it's a question of whether we want to present strings or DirectoryState.ACTIVE to the user
+    # leaving both options here for now until we settle on a preference
```
mattgd, line 51: Should this comment be removed now?

workos/directory_sync.py
```diff
@@ -34,13 +34,13 @@ def request_helper(self):
...
-        before=None,
-        after=None,
-        order=None,
-    ):
+        directory: Optional[str] = None,
+        group: Optional[str] = None,
+        limit: int = RESPONSE_LIMIT,
+        before: Optional[str] = None,
+        after: Optional[str] = None,
+        order: PaginationOrder = "desc",
```
tribble, line 42: `Optional[PaginationOrder] = "desc"` is not what we want here. That allows `None` as a possible value. This disallows `None` and does not require `order` to be specified.

workos/resources/organizations.py
```diff
@@ -1,29 +1,27 @@
...
+from typing import List, Literal, Optional, TypedDict
+from workos.resources.workos_model import WorkOSModel
 
 
-class WorkOSOrganization(WorkOSBaseResource):
-    """Representation of WorkOS Organization as returned by WorkOS through the Organizations feature.
+class OrganizationDomain(WorkOSModel):
+    id: str
+    organization_id: str
+    object: Literal["organization"]
```
mattgd, line 8: Should this be `Literal["organization_domain"]`?

workos/directory_sync.py
```diff
@@ -56,133 +56,42 @@ def list_users(
...
-                raise ValueError("Parameter order must be of enum type Order")
-
-        response = self.request_helper.request(
-            "directory_users",
-            method=REQUEST_METHOD_GET,
-            params=params,
-            token=workos.api_key,
+        return WorkOsListResource(
+            list_method=self.list_users,
+            # TODO: Should we even bother with this validation?
```
tribble, line 81: My current opinion is no. But want to constrain scope of this PR and fix later

workos/directory_sync.py
```diff
@@ -389,151 +208,40 @@ def list_directories(
...
-    def delete_directory(self, directory):
+    def delete_directory(self, directory: str):
         """Delete one existing Directory.
 
         Args:
             directory (str): The ID of the directory to be deleted. (Required)
 
         Returns:
-            dict: Directories response from WorkOS.
+            None
```
tribble, line 242: Yeah. Our doc here was a lie. We return an empty 202

workos/typing/enums.py
```diff
@@ -0,0 +1,48 @@
...
+    ValidationInfo,
+    ValidatorFunctionWrapHandler,
+    WrapValidator,
+)
+from workos.typing.untyped_literal import UntypedLiteral
+
+
+EnumType = TypeVar("EnumType", bound=Enum)
+
+# Identical to the literal or strings path, except meant for a literal of enum values.
```
tribble, line 16: Should I just delete this from the PR? Not being used at the moment.

workos/resources/list.py
```diff
@@ -90,3 +107,66 @@ def auto_paging_iter(self):
...
+
+
+class WorkOsListResource(BaseModel, Generic[T]):
+    object: Literal["list"]
+    data: List[T]
+    list_metadata: ListMetadata
+
+    # These fields end up exposed in the types. Does we care?
+    list_method: Callable = Field(exclude=True)
+    list_args: ListArgs = Field(exclude=True)
```
mattgd, line 144: Where do these fields this show up in types exactly?

workos/resources/list.py
```diff
@@ -90,3 +107,66 @@ def auto_paging_iter(self):
                 next_page_marker = response["list_metadata"][string_direction]
                 yield data
                 data = []
+
+
+# add all possible generics of List Resource
+T = TypeVar("T", Organization, Directory, DirectoryGroup, DirectoryUser)
```
mattgd, line 113: I feel like we could name this something better like "ListableResource".

workos/organizations.py
```diff
@@ -47,114 +50,30 @@ def list_organizations(
...
-
-        response = self.request_helper.request(
-            ORGANIZATIONS_PATH,
-            method=REQUEST_METHOD_GET,
-            params=params,
-            token=workos.api_key,
+        return WorkOsListResource[Organization](
+            list_method=self.list_organizations,
+            # TODO: Should we even bother with this validation?
+            list_args=ListArgs.model_validate(params),
```
mattgd, line 72: I feel like the call made Friday was that we wouldn't focus on inbound param validation. I'm okay with either, but we should probably be consistent.

workos/directory_sync.py
mattgd: Loaded up the page before you posted this, but down to omit. Better to be consistent everywhere IMO, and I think on Friday we decided against explicit input validation like this.

workos/typing/enums.py
mattgd: Yep, can delete.

workos/directory_sync.py
tribble: 👍 Will remove in a follow-on PR

workos/resources/organizations.py
tribble: It sure should

workos/organizations.py
tribble: Previously if `order` was not passed in, the value would fall back to `"desc"`.  Instead that default moves to the function signature: `order: PaginationOrder = "desc"`

workos/resources/list.py
tribble: They would appear a users IDE, like this
CleanShot 2024-07-22 at 13 29 42

workos/organizations.py
tribble: Will remove in a follow-up.

PR #285, opened 2024-07-18
Add asyncio support for events API methods

mattgd:
## Description
Add asyncio support for events API methods.

## Documentation

```
[ ] Yes
```


workos/__init__.py
```diff
@@ -1,9 +1,16 @@
...
-base_api_url = "https://api.workos.com/"
-request_timeout = 25
+base_api_url = os.getenv("WORKOS_BASE_URL")
+request_timeout = int(os.getenv("WORKOS_REQUEST_TIMEOUT", "25"))
+
+
+client = SyncClient(base_url=base_api_url, version=__version__, timeout=request_timeout)
+async_client = AsyncClient(
+    base_url=base_api_url, version=__version__, timeout=request_timeout
+)
```
mattgd: To prevent circular imports and promote client flexibility (such as for when we add FGA APIs), I pass in the base API URL. This means the base URL can't be updated on the fly with `workos.base_api_url = "new url"` like we do currently, so instead I'm pulling from an environment variable.

This feels pretty standard to me, but any thoughts?

workos/async_client.py
```diff
@@ -0,0 +1,77 @@
...
+from workos.sso import SSO
+from workos.webhooks import Webhooks
+from workos.mfa import Mfa
+from workos.events import AsyncEvents
+from workos.user_management import UserManagement
+
+from workos.utils.async_http_client import AsyncHTTPClient
+
+
+class AsyncClient(object):
```
mattgd, line 15: Instead of littering the SDK with `_events()` API methods, we'll have a separate AsyncClient, which developers can use methods from as needed.

workos/events.py
```diff
@@ -60,7 +55,73 @@ def list_events(
...
+            method=REQUEST_METHOD_GET,
+            params=params,
+            token=workos.api_key,
+        )
+
+        if "default_limit" in locals():
+            if "metadata" in response and "params" in response["metadata"]:
+                response["metadata"]["params"]["default_limit"] = default_limit
+            else:
+                response["metadata"] = {"params": {"default_limit": default_limit}}
```
mattgd, line 69: This should all be simplified with the auto paging changes in the other PR.

workos/__init__.py
```diff
@@ -1,9 +1,16 @@
...
-base_api_url = "https://api.workos.com/"
-request_timeout = 25
+base_api_url = os.getenv("WORKOS_BASE_URL", "https://api.workos.com/")
+request_timeout = int(os.getenv("WORKOS_REQUEST_TIMEOUT", "25"))
+
+
+client = SyncClient(base_url=base_api_url, version=__version__, timeout=request_timeout)
+async_client = AsyncClient(
+    base_url=base_api_url, version=__version__, timeout=request_timeout
+)
```
mattgd, line 16: To prevent circular imports and promote client flexibility (such as for when we add FGA APIs), I pass in the base API URL. This means the base URL can't be updated on the fly with `workos.base_api_url = "new url"` like we do currently, so instead I'm pulling from an environment variable.

This feels pretty standard to me, but any thoughts?

tribble: That seems OK. Could a developer just create an instance of the client themselves if they didn't want to use environment variables?

mattgd: That's a good point! They definitely could.

workos/utils/async_http_client.py
```diff
@@ -0,0 +1,96 @@
...
+            token (str): Bearer token
+
+        Returns:
+            dict: Response from WorkOS
+        """
+        url, headers, params, request_fn = self._prepare_request(
+            path=path, method=method, params=params, headers=headers, token=token
+        )
+
+        if method == REQUEST_METHOD_GET:
```
PaulAsjes, line 87: Do we only check for `GET` here? I don't think we use `PUT`, but we do use `DELETE` that shouldn't have a body.

PaulAsjes (changes requested):

workos/utils/async_http_client.py
mattgd: Fixed!

setup.py
```diff
@@ -27,17 +27,19 @@
     ),
     zip_safe=False,
     license=about["__license__"],
-    install_requires=["requests>=2.22.0"],
+    install_requires=["httpx>=0.27.0", "requests>=2.22.0"],
```
mattgd, line 30: `requests` will be phased out before we leave beta.

workos/utils/_base_http_client.py
```diff
@@ -0,0 +1,191 @@
...
+                "url": url,
+                "headers": headers,
+                "json": params,
+                "timeout": self.timeout,
+            }
+
+    def _handle_response(self, response: httpx.Response) -> dict:
+        response_json = None
+        content_type = (
+            response.headers.get("content-type")
```
tribble, line 147: So `response.headers` can be `None`? I guess that's possible, but unlikely.

workos/utils/_base_http_client.py
```diff
@@ -0,0 +1,191 @@
...
+            if response.headers is not None
+            else None
+        )
+        if content_type is not None and "application/json" in content_type:
+            try:
+                response_json = response.json()
+            except ValueError:
+                raise ServerException(response)
```
tribble, line 155: I normally think of server exceptions as 5xx. But content type not matching the content is pretty weird, so this is probably fine.

workos/utils/_base_http_client.py
```diff
@@ -0,0 +1,191 @@
...
+    def base_url(self, url: str) -> None:
+        """Creates an accessible template for constructing the URL for an API request.
+
+        Args:
+            base_api_url (str): Base URL for api requests
+        """
+        self._base_url = "{}{{}}".format(self._enforce_trailing_slash(url))
+
+    @property
+    def default_headers(self) -> "dict[str, str]":
```
tribble, line 175: When do you `"dict[str, str]"` in parseltongue and when do you `Dict[str, str]`?

workos/utils/http_client.py
```diff
@@ -0,0 +1,173 @@
...
+            base_url=base_url,
+            timeout=timeout,
+            follow_redirects=True,
+            transport=transport,
+        )
+
+    def is_closed(self) -> bool:
+        return self._client.is_closed
+
+    def close(self) -> None:
```
tribble, line 45: When does this get used?

workos/utils/http_client.py
```diff
@@ -0,0 +1,173 @@
...
+    """Async HTTP Client for a convenient way to access the WorkOS feature set."""
+
+    _client: httpx.AsyncClient
+
+    def __init__(
+        self,
+        *,
+        base_url: str,
+        version: str,
+        timeout: int = None,
```
tribble, line 112: Do you prefer to select a default time out or leave it to the underlying HTTP client default?

tribble (approved):
Nothing blocking.
Thumbs Up

tests/conftest.py
```diff
@@ -1,7 +1,11 @@
+from unittest.mock import AsyncMock, MagicMock
```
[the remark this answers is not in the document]
mattgd: I feel like I haven't seen Shia Labeouf in a bit.

tests/conftest.py
```diff
@@ -78,3 +82,23 @@ def capture_and_mock(*args, **kwargs):
         return (request_args, request_kwargs)
 
     return inner
+
+
+@pytest.fixture
+def mock_sync_http_client_with_response():
```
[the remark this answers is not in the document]
mattgd: Yeah that'd simplify things a bit. I'll see about working this (or something similar) in so we can generate the tests.

tests/test_async_events.py
```diff
@@ -0,0 +1,83 @@
...
+                    "organization_id": None,
+                    "after": None,
+                    "range_start": None,
+                    "range_end": None,
+                },
+                "method": AsyncEvents.list_events,
+            },
+        }
+
+    @pytest.mark.asyncio
```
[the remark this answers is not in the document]
mattgd: Annotations aren't really magic though, so there might be a way 👀

tests/test_async_http_client.py
```diff
@@ -0,0 +1,139 @@
...
+            return httpx.Response(200, json={"message": "Success!"})
+
+        self.http_client = AsyncHTTPClient(
+            base_url="https://api.workos.test",
+            version="test",
+            transport=httpx.MockTransport(handler),
+        )
+
+        self.http_client._client.request = AsyncMock(
+            return_value=httpx.Response(200, json={"message": "Success!"}),
```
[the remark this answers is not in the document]
mattgd: Oops, forgot to combine. Fixing!

tests/test_async_http_client.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+
+    @pytest.mark.asyncio
+    async def test_get_request(self):
+        response = await self.http_client.request(
+            "events", method="GET", params={"test_param": "test_value"}, token="test"
```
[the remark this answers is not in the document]
mattgd: `token` gets put in the `Authorization` header. We could probably simplify it a bit an have it part of the HTTP client logic, but kept it as-is for now since that's how it's been passed in the past.

tests/test_sync_http_client.py
```diff
@@ -0,0 +1,132 @@
...
+            return httpx.Response(200, json={"message": "Success!"})
+
+        self.http_client = SyncHTTPClient(
+            base_url="https://api.workos.test",
+            version="test",
+            transport=httpx.MockTransport(handler),
+        )
+
+        self.http_client._client.request = MagicMock(
+            return_value=httpx.Response(200, json={"message": "Success!"}),
```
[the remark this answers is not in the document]
mattgd: Good call. I'll add a test!

workos/async_client.py
```diff
@@ -0,0 +1,68 @@
...
+
+    _http_client: AsyncHTTPClient
+
+    def __init__(self, base_url: str, version: str, timeout: int):
+        self._http_client = AsyncHTTPClient(
+            base_url=base_url, version=version, timeout=timeout
+        )
+
+    @property
+    def sso(self):
```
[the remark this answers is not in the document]
mattgd: Yep, they should. I can make that change.

workos/utils/_base_http_client.py
```diff
@@ -0,0 +1,191 @@
...
+    def _build_headers(
+        self, custom_headers: Union[dict, None], token: str = None
+    ) -> httpx.Headers:
+        if custom_headers is None:
+            custom_headers = {}
+
+        if token:
+            custom_headers["Authorization"] = "Bearer {}".format(token)
+
+        # httpx.Headers is case-insensitive while dictionaries are not.
```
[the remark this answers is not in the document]
mattgd: Just an FYI 👍

workos/utils/_base_http_client.py
mattgd: Brought this over from `requests`. Let me see what httpx says.

workos/utils/_base_http_client.py
mattgd: I think this is Python versioning differences. I'll change to Dict[str, str] to match our other type hints, but in newer versions of Python there is no difference it's just a quality of life improvement (no import).

workos/utils/_base_http_client.py
```diff
@@ -0,0 +1,191 @@
...
+        self._timeout = timeout
+
+    def _enforce_trailing_slash(self, url: str) -> str:
+        return url if url.endswith("/") else url + "/"
+
+    def _generate_api_url(self, path: str) -> str:
+        return self._base_url.format(path)
+
+    def _build_headers(
+        self, custom_headers: Union[dict, None], token: str = None
```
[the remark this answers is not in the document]
mattgd: This is a bit of a Python gotcha. The default value won't be used if we pass the `custom_headers` parameter at all (even if it's None). It's only used when `custom_headers` is not passed.

workos/utils/http_client.py
mattgd: Took this from Stainless. We don't use close directly, but it is used if the `__exit__` method is called by the Python runtime.  This is mostly a helper if we want to cleanup on a fatal error.

workos/utils/http_client.py
mattgd: Good question. We have a configurable option in our current Python SDK version, so I added. Not sure how many people use it though.

tests/conftest.py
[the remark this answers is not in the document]
tribble: Shia, The Beef.

workos/utils/_base_http_client.py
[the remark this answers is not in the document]
tribble: I'm not sure I'm understanding the gotcha.

>  The default value won't be used if we pass the custom_headers parameter at all

...that seems OK? Overriding the default is what I'd expect when passing in a param.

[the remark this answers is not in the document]
mattgd: We always want it to end up as a dict, not None, before we tack on our base headers. Otherwise will get a NoneType is not subscriptable error.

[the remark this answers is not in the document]
tribble: Right, that's why I suggested removing the `None` and just making the type `custom_headers: Dict = {}`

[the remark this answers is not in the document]
mattgd: Ah, sorry let me give an example. I don't think the type hinting is strong enough to prevent a `None` value disguised as a `dict`:


```python
def test_with_default(value: str = "test"):
    return f"[{value}]"
    
def test_without_default(value: str):
    return f"[{value}]"
    
def test_with_default_in_function(value: str):
    return f"[{value if value is not None else 'test'}]"
    
print(f"When test_with_default is given a value: {test_with_default('provided value')}")
print(f"When test_without_default is given a value: {test_without_default('provided value')}")
print(f"When test_with_default_in_function is given a value: {test_with_default_in_function('provided value')}")

print(f"When test_with_default is given a value kwarg: {test_with_default(value='provided value')}")
print(f"When test_without_default is given a value kwarg: {test_without_default(value='provided value')}")
print(f"When test_with_default_in_function is given a value kwarg: {test_with_default_in_function(value='provided value')}")

print(f"When test_with_default is NOT given a value: {test_with_default()}")
# Throws error: print(f"When test_without_default is NOT given a value: {test_without_default()}")
# Throws error: print(f"When test_with_default_in_function is NOT given a value: {test_with_default_in_function()}")

# This is the case we want to prevent None for. Without the extra check we'd get customer_headers = None since
# we always pass custom_headers:
value = None
print(f"When test_without_default is given a None value: {test_with_default(value=value)}")
print(f"When test_without_default is given a None value: {test_without_default(value=value)}")
print(f"When test_without_default is given a None value: {test_with_default_in_function(value=value)}")
```

Output:
```
When test_with_default is given a value: [provided value]
When test_without_default is given a value: [provided value]
When test_with_default_in_function is given a value: [provided value]
When test_with_default is given a value kwarg: [provided value]
When test_without_default is given a value kwarg: [provided value]
When test_with_default_in_function is given a value kwarg: [provided value]
When test_with_default is NOT given a value: [test]
When test_without_default is given a None value: [None]
When test_without_default is given a None value: [None]
When test_without_default is given a None value: [test]
```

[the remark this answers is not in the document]
tribble: And you don't see this in the type hinting?
CleanShot 2024-07-23 at 11 37 18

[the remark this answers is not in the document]
gcarvelli: Just a heads up: it's best practice to not use objects as default param values in Python because Python only instantiates them once: https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python. Strings/ints/other primitives are okay.

It's goofy and annoying to work around. I think your options are
1. Require a dict to be passed in, no default
2. Require Union[dict, None] to be passed in and instantiate if None (which feels like the behavior you want here). You could even default to `None` to make the parameter fully optional, if that's not against the refactor strategy you're using. That's a pattern I've seen in libraries I've used in the past.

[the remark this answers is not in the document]
tribble: Oh. Well that's horrible. OK, `None` it is.

But just to be clear, it's only a problem if we mutate that default value or return it such that it could be mutated later, correct?

PR #287, opened 2024-07-22
List typing fixes

tribble:
## Description
Rather than check list params at runtime, just define static types. I had to refactor the list resource a bit to accomplish this, as I moved away from using a model for the args and just used a `TypedDict`

## Documentation

```
[ ] Yes
```


workos/resources/list.py
```diff
@@ -117,56 +124,55 @@ def auto_paging_iter(self):
...
+    Generic[ListableResource, FilterParams],
+):
     object: Literal["list"]
     data: List[ListableResource]
     list_metadata: ListMetadata
 
-    # These fields end up exposed in the types. Does we care?
     list_method: Callable = Field(exclude=True)
     list_args: ListArgs = Field(exclude=True)
+    filter_params: FilterParams = Field(exclude=True)
```
tribble, line 158: I don't love passing in an entirely separate type for this, but guess what? Python type hinting doesn't support intersection between two `TypedDict`s. This is why we can't have nice things.

https://github.com/python/typing/discussions/1275

tribble: ... actually, I wonder if all of the list resource params can inherit from `ListArgs` and add their own specific params. I think that will work out better. Going to convert this back to drafe.

mattgd (approved):

PR #288, opened 2024-07-24
Add async support for Directory Sync

mattgd:
## Description
Add async support for DirectorySync.

## Documentation

```
[ ] Yes
```


workos/resources/list.py
```diff
@@ -178,3 +197,35 @@ def auto_paging_iter(self) -> Iterator[ListableResource]:
...
+                else:
+                    return
+            yield self.data[index]
+            index += 1
+
+
+SyncOrAsyncListResource = Union[
+    Awaitable[AsyncWorkOsListResource],
+    WorkOsListResource,
+]
```
mattgd, line 231: So this is FUN. I wanted this to be:
```
SyncOrAsyncListResource = Union[
    Awaitable[AsyncWorkOsListResource[ListableResource, ListAndFilterParams]],
    WorkOsListResource[ListableResource, ListAndFilterParams],
]
```
but Unions cannot be used as generics. This means I had to make the `Protocol` return type `SyncOrAsyncListResource` for list methods, instead of something like `SyncOrAsyncListResource[Directory, DirectoryFilterParams]`.

workos/directory_sync.py
```diff
@@ -297,7 +297,229 @@ def list_directories(
...
+            method=REQUEST_METHOD_GET,
+            params=list_params,
+            token=workos.api_key,
+        )
+
+        return AsyncWorkOsListResource(
+            list_method=self.list_users,
+            list_args=list_params,
+            **ListPage[DirectoryUser](**response).model_dump(),
+        )
```
mattgd, line 373: This PR lacks deduplication of code like this. Might be a good thing for me to take a stab at either before or after I work on connections.

workos/resources/list.py
tribble: I think to do something similar for events I'm creating a union of all event types and using that as a `bounds=` constraint on a type variable.... I don't know if that's helpful at all.

workos/utils/_base_http_client.py
```diff
@@ -32,8 +33,8 @@ class PreparedRequest(TypedDict):
     method: str
     url: str
     headers: httpx.Headers
-    params: NotRequired[Union[Dict, None]]
-    json: NotRequired[Union[Dict, None]]
+    params: NotRequired[Union[Mapping, None]]
```
tribble, line 36: Oh. I just did the same thing in my PR. That's fine, git diffing should be good about any conflicts

mattgd: Yeah, I figured it should be okay. Just wanted to get my tests passing.

workos/resources/list.py
mattgd: Let me give that a go.

PR #289, opened 2024-07-24
Add Event types

tribble:
## Description
Initial PR to add two event types. It's most crucial to get feedback on the approach, then we can add more types in following PRs.
## Documentation

```
[ ] Yes
```


workos/events.py
```diff
@@ -58,17 +74,16 @@ def list_events(
...
+        params: EventsListFilters = {
             "events": events,
             "limit": limit,
             "after": after,
-            "organization_id": organization_id,
+            "organization_id": organization,
             "range_start": range_start,
             "range_end": range_end,
+            # TODO: This is a hack, and it's wrong. Events does not support before or order
+            "before": None,
```
mattgd, line 85: I'm a bit unclear - why do we need to define `before` here?

workos/events.py
```diff
@@ -145,15 +150,8 @@ async def list_events(
...
-            "params": params,
-            "method": AsyncEvents.list_events,
-        }
-
-        return response
+        return WorkOsListResource(
+            list_method=self.list_events,
+            list_args=params,
+            **ListPage[Event](**response).model_dump(),
+        )
```
mattgd, line 157: Oh interesting. Does this work without issue for async? I may have to revisit this in my PR.

workos/events.py
tribble: Our current list params typing forces it. We need to refactor but I'm kicking that can down the road to a future PR.

workos/events.py
tribble: I doubt it. We probably need to parametrize the list resource to use an async client. But I haven't gotten there yet

tribble:
@mattgd Switching this back to draft. Need to fix some tests and there may be some real work to make the auto paging iter async

mattgd:
> @mattgd Switching this back to draft. Need to fix some tests and there may be some real work to make the auto paging iter async

Before you troubleshoot, I have an addition for that in my PR - let me know what you think: https://github.com/workos/workos-python/pull/288/files#diff-fd823a1e532828a74cb86c58012687832231229d0985952f4580c40ab33c3191R206-R225.

tribble:
@mattgd OK, PR ready for review. Events has enough quirks this required some changes. Will comment inline.

tests/conftest.py
```diff
@@ -153,7 +153,7 @@ def inner(
         http_client: Union[SyncHTTPClient, AsyncHTTPClient],
         data_list: list,
         status_code: int = 200,
-        headers: Mapping[str, str] = None,
+        headers: Optional[Mapping[str, str]] = None,
```
tribble, line 156: Maybe we should point mypy at more of our files?

tests/test_directory_sync.py
```diff
@@ -8,6 +8,20 @@
 from tests.utils.fixtures.mock_directory_group import MockDirectoryGroup
 
 
+def api_directory_to_sdk(directory):
+    # The API returns an active directory as 'linked'
+    # We normalize this to 'active' in the SDK. This helper function
+    # does this conversion to make make assertions easier.
+    if directory["state"] == "linked":
+        return {**directory, "state": "active"}
```
tribble, line 16: Yeah, I bit the bullet and just did this conversion instead of having two types.

tests/test_directory_sync.py
```diff
@@ -73,7 +87,7 @@ def mock_group(self):
 
     @pytest.fixture
     def mock_directories(self):
-        directory_list = [MockDirectory(id=str(i)).to_dict() for i in range(20)]
+        directory_list = [MockDirectory(id=str(i)).to_dict() for i in range(10)]
```
tribble, line 90: Our list pages default to 10 anyway

tests/utils/fixtures/mock_directory.py
```diff
@@ -8,8 +8,9 @@ def __init__(self, id):
         self.object = "directory"
         self.id = id
         self.organization_id = "organization_id"
-        self.domain = "crashlanding.com"
-        self.name = "Ri Jeong Hyeok"
+        self.external_key = "ext_123"
```
tribble, line 11: This external key is actually required. See our [API ref](https://workos.com/docs/reference/directory-sync/directory) (which isn't consistent, *sigh*) as well as our [API JSON model](https://github.com/workos/workos/blob/e084d3669e08a2eda7d3e8b4911668605fb9ead1/packages/api-json-models/src/rest-api/directory.ts#L10) and [Event JSON model](https://github.com/workos/workos/blob/8c93abedf98f81584088b56329dec0de9488e21f/packages/api-json-models/src/events/shared/directory.ts#L66).

tests/utils/fixtures/mock_directory_activated_payload.py
```diff
@@ -0,0 +1,30 @@
+import datetime
+from workos.resources.base import WorkOSBaseResource
+
+
+class MockDirectoryActivatedPayload(WorkOSBaseResource):
```
tribble, line 5: This was easier than mangling the existing `MockDirectory` model

tests/utils/fixtures/mock_event.py
```diff
@@ -1,13 +1,17 @@
...
 from workos.resources.base import WorkOSBaseResource
 
 
 class MockEvent(WorkOSBaseResource):
     def __init__(self, id):
         self.object = "event"
         self.id = id
-        self.event = "dsync.user.created"
-        self.data = {"id": "event_01234ABCD", "organization_id": "org_1234"}
+        self.event = "dsync.activated"
```
tribble, line 13: Had to switch to an event that actually exists

tests/utils/list_resource.py
```diff
@@ -1,7 +1,7 @@
-from typing import Dict, List, Optional
+from typing import Dict, Optional, Sequence
 
 
-def list_data_to_dicts(list_data: List):
+def list_data_to_dicts(list_data: Sequence):
```
tribble, line 4: Why not use a more generic duck type?

workos/directory_sync.py
```diff
@@ -4,7 +4,6 @@
 from workos.utils.http_client import AsyncHTTPClient, SyncHTTPClient
 from workos.utils.pagination_order import PaginationOrder
 from workos.utils.request import REQUEST_METHOD_DELETE, REQUEST_METHOD_GET
-
```
tribble, line 7: I hates me rando newlines

workos/events.py
```diff
@@ -57,15 +70,11 @@ def list_events(
...
-            limit = RESPONSE_LIMIT
-            default_limit = True
-
-        params = {
+        params: EventsListFilters = {
             "events": events,
             "limit": limit,
             "after": after,
-            "organization_id": organization_id,
+            "organization_id": organization,
```
tribble, line 77: Right now our other [API methods](https://github.com/workos/workos-python/blob/79a5de376d4645e28a3005356a15fe1a61eb35b1/workos/organizations.py#L116) just use `organization`.

I'm conforming for consistency. Happy to have a cleanup PR sweep through and change all of those params to `organization_id`

workos/events.py
```diff
@@ -76,22 +85,14 @@ def list_events(
...
-        response["metadata"] = {
-            "params": params,
-            "method": Events.list_events,
-        }
-
-        return response
+        return WorkOsListResource(
+            list_method=self.list_events,
+            list_args=params,
+            **ListPage[Event](**response).model_dump(exclude_unset=True),
```
tribble, line 91: I had to use `exclude_unset=True` otherwise this would always serialize `before` in the `list_metadata`. I think it's because the resource can't tell the difference between `ListAfterMetadata` and `ListMetadata`.

The other option is to and a third generic parameter and explicitly pass in the `ListMetadata` shape. But, since python type hint generics don't set support default values for generics, every list resource would have to pass in the list metadata type.

workos/resources/directory_sync.py
```diff
@@ -1,15 +1,8 @@
 from typing import List, Optional, Literal
 from workos.resources.workos_model import WorkOSModel
+from workos.types.directory_sync.directory_state import DirectoryState
 from workos.typing.literals import LiteralOrUntyped
 
-DirectoryState = Literal[
```
tribble, line 5: Extracted to a shared type which normalizes `state`

workos/resources/event.py
```diff
@@ -1,42 +0,0 @@
-from workos.resources.base import WorkOSBaseResource
```
tribble, line 1: Wasn't used ¯\\\_(ツ)_/¯

workos/resources/list.py
```diff
@@ -147,16 +156,22 @@ class BaseWorkOsListResource(
...
-            "order": self.list_args["order"],
-            "limit": self.list_args["limit"],
-        }
+        fixed_pagination_params = cast(
+            # Type hints consider this a mismatch because it assume the dictionary is dict[str, int]
+            Dict[str, Union[int, str, None]],
+            {
+                "limit": self.list_args["limit"],
+            },
+        )
```
tribble, line 171: Type hints were a bit "too smart". Since `self.list_args["limit"] is an `int` it considers the entire dictionary to be of type `Dict[str, int]`.

workos/resources/list.py
```diff
@@ -147,16 +156,22 @@ class BaseWorkOsListResource(
...
+        if "order" in self.list_args:
```
tribble, line 172: Order isn't a thing for events.

workos/events.py
tribble: I realized there is another option. We could have separate `WorkOsForwardListResource` and `WorkOsBidirectionalListResource`.

Not sure how much benefit that will bring. It would almost definitely be more code.

workos/types/directory_sync/directory_state.py
```diff
@@ -0,0 +1,25 @@
...
+    "deleting",
+    "invalid_credentials",
+]
+
+
+def convert_linked_to_active(value: Any, info: ValidationInfo) -> Any:
+    if isinstance(value, str) and value == "linked":
+        return "active"
+    else:
+        return value
```
mattgd, line 19: 🙃

tests/conftest.py
mattgd: Seemingly we should be able to open it up to all `workos/` files. That's how I was running it locally. If we're having a mypy mismatch we should adjust settings to make sure it runs how we'd like in CI.

mattgd (approved):

PR #291, opened 2024-07-25
Add typing and async support for SSO

mattgd:
## Description
Add typing and async support for SSO.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/tests/test_sso.py
+++ b/tests/test_sso.py
@@ -82,7 +82,7 @@ class TestSSOBase(SSOFixtures):
         self.customer_domain = "workos.com"
         self.login_hint = "foo@workos.com"
         self.redirect_uri = "https://localhost/auth/callback"
-        self.state = json.dumps({"things": "with_stuff"})
+        self.authorization_state = json.dumps({"things": "with_stuff"})
         self.connection_id = "connection_123"
         self.organization_id = "organization_123"
         self.setup_completed = True
@@ -92,22 +92,25 @@ class TestSSOBase(SSOFixtures):
     ):
         with pytest.raises(ValueError, match=r"Incomplete arguments.*"):
             self.sso.get_authorization_url(
-                redirect_uri=self.redirect_uri, state=self.state
+                redirect_uri=self.redirect_uri, state=self.authorization_state
             )
 
     def test_authorization_url_has_expected_query_params_with_provider(self):
         authorization_url = self.sso.get_authorization_url(
-            provider=self.provider, redirect_uri=self.redirect_uri, state=self.state
+            provider=self.provider,
+            redirect_uri=self.redirect_uri,
+            state=self.authorization_state,
         )
 
         parsed_url = urlparse(authorization_url)
+        print(parsed_url, authorization_url)
 
         assert dict(parse_qsl(parsed_url.query)) == {
             "provider": self.provider,
             "client_id": workos.client_id,
             "redirect_uri": self.redirect_uri,
             "response_type": RESPONSE_TYPE_CODE,
-            "state": self.state,
+            "state": self.authorization_state,
         }
 
     def test_authorization_url_has_expected_query_params_with_domain_hint(self):
@@ -115,7 +118,7 @@ class TestSSOBase(SSOFixtures):
             connection_id=self.connection_id,
             domain_hint=self.customer_domain,
             redirect_uri=self.redirect_uri,
-            state=self.state,
+            state=self.authorization_state,
         )
 
         parsed_url = urlparse(authorization_url)
@@ -126,7 +129,7 @@ class TestSSOBase(SSOFixtures):
             "redirect_uri": self.redirect_uri,
             "connection": self.connection_id,
             "response_type": RESPONSE_TYPE_CODE,
-            "state": self.state,
+            "state": self.authorization_state,
         }
 
     def test_authorization_url_has_expected_query_params_with_login_hint(self):
@@ -134,7 +137,7 @@ class TestSSOBase(SSOFixtures):
             connection_id=self.connection_id,
             login_hint=self.login_hint,
             redirect_uri=self.redirect_uri,
-            state=self.state,
+            state=self.authorization_state,
         )
 
         parsed_url = urlparse(authorization_url)
@@ -145,14 +148,14 @@ class TestSSOBase(SSOFixtures):
             "redirect_uri": self.redirect_uri,
             "connection": self.connection_id,
             "response_type": RESPONSE_TYPE_CODE,
-            "state": self.state,
+            "state": self.authorization_state,
         }
 
     def test_authorization_url_has_expected_query_params_with_connection(self):
         authorization_url = self.sso.get_authorization_url(
             connection_id=self.connection_id,
             redirect_uri=self.redirect_uri,
-            state=self.state,
+            state=self.authorization_state,
         )
 
         parsed_url = urlparse(authorization_url)
@@ -162,7 +165,7 @@ class TestSSOBase(SSOFixtures):
             "client_id": workos.client_id,
             "redirect_uri": self.redirect_uri,
             "response_type": RESPONSE_TYPE_CODE,
-            "state": self.state,
+            "state": self.authorization_state,
         }
 
     def test_authorization_url_with_string_provider_has_expected_query_params_with_organization(
@@ -172,7 +175,7 @@ class TestSSOBase(SSOFixtures):
             provider=self.provider,
             organization_id=self.organization_id,
             redirect_uri=self.redirect_uri,
-            state=self.state,
+            state=self.authorization_state,
         )
 
         parsed_url = urlparse(authorization_url)
@@ -183,14 +186,14 @@ class TestSSOBase(SSOFixtures):
             "client_id": workos.client_id,
             "redirect_uri": self.redirect_uri,
             "response_type": RESPONSE_TYPE_CODE,
-            "state": self.state,
+            "state": self.authorization_state,
         }
 
     def test_authorization_url_has_expected_query_params_with_organization(self):
         authorization_url = self.sso.get_authorization_url(
             organization_id=self.organization_id,
             redirect_uri=self.redirect_uri,
-            state=self.state,
+            state=self.authorization_state,
         )
 
         parsed_url = urlparse(authorization_url)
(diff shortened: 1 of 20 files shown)


tests/test_sso.py
```diff
@@ -417,190 +268,231 @@ def test_get_profile_and_token_returns_expected_workosprofile_object(
...
+        all_connections = []
 
-    def test_delete_connection(self, setup_with_client_id, mock_raw_request_method):
-        mock_raw_request_method(
-            "delete",
-            "No Content",
-            204,
-            headers={"content-type": "text/plain; charset=utf-8"},
+        for connection in connections.auto_paging_iter():
+            all_connections.append(connection)
```
tribble, line 354: We repeat something like this for every auto paging test. I wonder if it's worth having some helper. (Not in this PR)

tests/test_sso.py
```diff
@@ -417,190 +268,231 @@ def test_get_profile_and_token_returns_expected_workosprofile_object(
...
+            response_dict=mock_connections,
+            status_code=200,
+        )
 
-        assert len(*list(all_connections)) == len(
-            dict_mock_connections_with_limit["data"]
+        await self.sso.list_connections(connection_type="GenericSAML")
+
+        request_params = request_kwargs["params"]
+        assert request_params["connection_type"] == "GenericSAML"
```
tribble, line 465: For tests like this I wonder if we should explicitly assert all the params, like `assert {"connection_type": "GenericSaml", "limit": 10} == request_kwargs["params"]` or something similar. We know exactly how we're calling `list_connections`. it seems like we should know exactly how that translates into the HTTP request.

tests/utils/fixtures/mock_connection.py
```diff
@@ -4,16 +4,21 @@
 
 class MockConnection(WorkOSBaseResource):
     def __init__(self, id):
-        self.object = "organization"
```
tribble, line 7: 🤦

workos/resources/list.py
```diff
@@ -116,11 +117,12 @@ def auto_paging_iter(self):
 ListableResource = TypeVar(
     # add all possible generics of List Resource
     "ListableResource",
-    Organization,
+    Connection,
```
tribble, line 120: Alphabetical is a good idea.

workos/resources/sso.py
```diff
@@ -1,87 +1,59 @@
...
-        "connection_id",
-        "connection_type",
-        "idp_id",
-        "raw_attributes",
-    ]
+    object: Literal["profile"]
+    id: str
+    connection_id: str
+    connection_type: LiteralOrUntyped[ConnectionType]
+    organization_id: Union[str, None]
```
tribble, line 15: I had no idea it was possible to have a `Profile` or `Connection` with no `organization_id`

tribble: Oh. OAuth. That's how.

workos/resources/sso.py
```diff
@@ -1,87 +1,59 @@
...
-class WorkOSProfileAndToken(WorkOSBaseResource):
-    """Representation of a User Profile and Access Token as returned by WorkOS through the SSO feature.
+class ProfileAndToken(WorkOSModel):
+    """Representation of a User Profile and Access Token as returned by WorkOS through the SSO feature."""
 
-    Attributes:
-        OBJECT_FIELDS (list): List of fields a WorkOSProfileAndToken is comprised of.
-    """
+    access_token: str
+    profile: Profile
```
tribble, line 28: Oh. Cute. Another of these combo return types.

workos/resources/sso.py
```diff
@@ -1,87 +1,59 @@
...
-    OBJECT_FIELDS = [
-        "access_token",
-    ]
 
-    @classmethod
-    def construct_from_response(cls, response):
-        profile_and_token = super(WorkOSProfileAndToken, cls).construct_from_response(
-            response
-        )
+ConnectionState = Literal["linked", "unlinked"]
```
tribble, line 31: OK, our [API reference](https://workos.com/docs/reference/sso/connection) claim the valid values are `active`, `inactive`, and `validating`. But looking at the actual [JSON model](https://github.com/workos/workos/blob/9b93b7ad269ec28a4a0f9d024c84475cd44c5406/packages/api-json-models/src/shared/connection-state.ts#L3), we have 6 different states.

I assume some of these states might always be filtered out by the service. But if they're impossible to expose, they should not be in our API JSON model. :/

workos/sso.py
```diff
@@ -1,116 +1,71 @@
...
-        login_hint=None,
-        redirect_uri=None,
-        state=None,
-        provider=None,
-        connection=None,
-        organization=None,
-    ) -> str: ...
-
-    def get_profile(self, accessToken: str) -> WorkOSProfile: ...
+DEFAULT_RESPONSE_LIMIT = 10
```
tribble, line 35: Ever wonder why we don't create one constant for this? Is it a python thing?

workos/sso.py
```diff
@@ -121,69 +76,83 @@ def get_authorization_url(
...
-            "GET",
-            self.request_helper.generate_api_url(AUTHORIZATION_PATH),
-            params=params,
-        ).prepare()
+        request_url = fake_http_client.build_request_url(
+            url=AUTHORIZATION_PATH, method=REQUEST_METHOD_GET, params=params
+        )
 
-        return prepared_request.url
+        fake_http_client.close()
```
tribble, line 107: ... this is the idiomatic way to construct a URL? That's truly Slytherin

tribble: [Is `urlencode`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) out of favor?

workos/sso.py
```diff
@@ -248,80 +217,120 @@ def list_connections(
...
-            limit = RESPONSE_LIMIT
-            default_limit = True
 
         params = {
-            "connection_type": connection_type.value if connection_type else None,
+            "connection_type": connection_type,
             "domain": domain,
             "organization_id": organization_id,
-            "limit": limit,
+            "limit": limit or DEFAULT_RESPONSE_LIMIT,
```
tribble, line 225: Didn't this already get a default above?

workos/sso.py
```diff
@@ -248,80 +217,120 @@ def list_connections(
...
-            except KeyError:
-                raise ValueError("'connection_type' must be a member of ConnectionType")
-
-        if limit is None:
-            limit = RESPONSE_LIMIT
-            default_limit = True
 
         params = {
-            "connection_type": connection_type.value if connection_type else None,
+            "connection_type": connection_type,
```
tribble, line 222: We are not consistent about when we always add a key (even if it's `None`) or only add the key to the params if it's specified. I'm pretty sure I'm guilty of this as well. We should pick a style and go with it.

workos/sso.py
```diff
@@ -248,80 +217,120 @@ def list_connections(
...
             token=workos.api_key,
         )
 
-        response["metadata"] = {
-            "params": params,
-            "method": SSO.list_connections,
+        return WorkOsListResource(
+            list_method=self.list_connections,
+            list_args=params,
+            **ListPage[Connection](**response).model_dump(exclude_unset=True),
```
tribble, line 241: I explicitly added `exclude_unset` to the events endpoint because I didn't want to serialize the `before` key, which does not exist in the event list metadata.

Do you think we should extend this to all list endpoints? It might be moot, the API may not omit any fields.

workos/sso.py
```diff
@@ -336,73 +345,37 @@ def list_connections_v2(
...
-            limit = RESPONSE_LIMIT
-            default_limit = True
-
         params = {
-            "connection_type": connection_type.value if connection_type else None,
+            "connection_type": connection_type,
             "domain": domain,
             "organization_id": organization_id,
-            "limit": limit,
+            "limit": limit or DEFAULT_RESPONSE_LIMIT,
```
tribble, line 352: Same here, didn't this get a value above?

workos/utils/connection_types.py
```diff
@@ -1,48 +1,39 @@
...
+    "OracleSAML",
+    "PingFederateSAML",
+    "PingOneSAML",
+    "RipplingSAML",
+    "SalesforceSAML",
+    "ShibbolethGenericSAML",
+    "ShibbolethSAML",
+    "SimpleSamlPhpSAML",
+    "VMwareSAML",
+]
```
tribble, line 39: The API types also include [`pending`](https://github.com/workos/workos/blob/a91712330d5e55492c79ef71c480eb27497d750b/packages/api-json-models/src/shared/connection-type.ts#L7) and [`Auth0Migration`](https://github.com/workos/workos/blob/a91712330d5e55492c79ef71c480eb27497d750b/packages/api-json-models/src/shared/connection-type.ts#L12) types.

I think whoever owns SSO (is that us? LOL) could give us guidance around whether it's safe to omit these.

workos/utils/http_client.py
```diff
@@ -172,3 +172,26 @@ async def request(
         )
         response = await self._client.request(**prepared_request_parameters)
         return self._handle_response(response)
+
+
+class FakeHTTPClient(SyncHTTPClient):
```
tribble, line 177: 🙈

tribble (commented):
See comments inline. Creating an HTTP client to generate a URL seems like such a smell. Is there a better way?

workos/sso.py
tribble: ... and I realized, if we do want to omit missing keys, I wonder if it's easier to come full circle and define our params as `BaseModel`s.

Then we could just....
```python
params = ConnectionParams(connection_type=connection_type)
# when it's time to use the params
Request(params=ConnectionParams.model_dump(exclude_unset=True))
```

tests/test_sso.py
```diff
@@ -63,218 +56,63 @@ def mock_connection(self):
...
+    def setup(self, set_api_key_and_client_id):
+        self.http_client = SyncHTTPClient(
+            base_url="https://api.workos.test", version="test"
+        )
+        self.sso = SSO(http_client=self.http_client)
+        self.provider = "GoogleOAuth"
+        self.customer_domain = "workos.com"
+        self.login_hint = "foo@workos.com"
+        self.redirect_uri = "https://localhost/auth/callback"
+        self.state = json.dumps({"things": "with_stuff"})
```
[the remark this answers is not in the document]
tribble: Ah. Damn you `state` and how general you are.

workos/sso.py
tribble: Oh. In that case I guess we don't need to care.

tribble (approved):
Sheep it
sheep

PR #292, opened 2024-07-25
Type all dsync events

tribble:
## Description
Add types for the rest of the Dsync events. Needed to refactor the API return type. The Event does not contain groups.

## Documentation

```
[ ] Yes
```


workos/_base_client.py
```diff
@@ -19,31 +19,41 @@ class BaseClient(Protocol):
     _http_client: BaseHTTPClient
 
     @property
-    def audit_logs(self) -> AuditLogsModule: ...
+    def audit_logs(self) -> AuditLogsModule:
```
tribble, line 22: ... do we have different configs for `black`? Or different versions?

mattgd: I updated the version of black in the repo. I believe if you try reinstalling the version matching the setup.py you, me and CI will be in sync.

tribble: Yup, needed to `pip install -U black`

workos/types/directory_sync/directory_user.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class DirectoryUserEmail(WorkOSModel):
+    type: Optional[str] = None
+    value: Optional[str] = None
+    primary: Optional[bool] = None
+
+
+class Role(WorkOSModel):
+    slug: str
```
mattgd, line 16: We could reuse this for org. memberships, so I may end up moving it.

mattgd (approved):

workos/types/directory_sync/directory_user.py
tribble: Yup, I figured it might move around

workos/resources/events.py
```diff
@@ -39,7 +68,66 @@ class DirectoryDeletedEvent(EventModel[Literal["dsync.deleted"], DirectoryPayloa
     event: Literal["dsync.deleted"]
 
 
+class DirectoryGroupCreatedEvent(
+    EventModel[Literal["dsync.group.created"], DirectoryGroup]
+):
+    event: Literal["dsync.group.created"]
```
tribble, line 74: When I set `pylance` to standard type checking rather than basic, it doesn't like this property override because it doesn't match the type on `EventModel`.

I'm going to ignore it for now, but it may rear its head again when I test what happens when an unrecognized event type gets parsed.

PR #294, opened 2024-07-26
Add typing for User Management

mattgd:
## Description
Add typing for User Management.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,12 +1,13 @@
-from typing import Mapping, Optional, Union
+from typing import Any, Callable, Mapping, Optional, Union
 from unittest.mock import AsyncMock, MagicMock
 
 import httpx
 import pytest
 import requests
 
-from tests.utils.list_resource import list_response_of
+from tests.utils.list_resource import list_data_to_dicts, list_response_of
 import workos
+from workos.resources.list import WorkOsListResource
 from workos.utils.http_client import AsyncHTTPClient, SyncHTTPClient
 
 
@@ -149,15 +150,13 @@ def mock_http_client_with_response(monkeypatch):
 def capture_and_mock_http_client_request(monkeypatch):
     def inner(
         http_client: Union[SyncHTTPClient, AsyncHTTPClient],
-        response_dict: dict,
+        response_dict: Optional[dict] = None,
         status_code: int = 200,
         headers: Optional[Mapping[str, str]] = None,
     ):
-        request_args = []
         request_kwargs = {}
 
         def capture_and_mock(*args, **kwargs):
-            request_args.extend(args)
             request_kwargs.update(kwargs)
 
             return httpx.Response(
@@ -173,7 +172,7 @@ def capture_and_mock_http_client_request(monkeypatch):
 
         monkeypatch.setattr(http_client._client, "request", mock)
 
-        return (request_args, request_kwargs)
+        return request_kwargs
 
     return inner
 
@@ -222,5 +221,33 @@ def mock_pagination_request_for_http_client(monkeypatch):
         mock = mock_class(side_effect=mock_function)
 
         monkeypatch.setattr(http_client._client, "request", mock)
+
+    return inner
+
+
+@pytest.fixture
+def test_sync_auto_pagination(
+    mock_pagination_request_for_http_client,
+):
+    def inner(
+        http_client: Union[SyncHTTPClient, AsyncHTTPClient],
+        list_function: Callable[[], WorkOsListResource],
+        expected_all_page_data: dict,
+        list_function_params: Optional[Mapping[str, Any]] = {},
+    ):
+        mock_pagination_request_for_http_client(
+            http_client=http_client,
+            data_list=expected_all_page_data,
+            status_code=200,
+        )
+
+        results = list_function(**list_function_params or {})
+        all_results = []
+
+        for result in results.auto_paging_iter():
+            all_results.append(result)
+
+        assert len(list(all_results)) == len(expected_all_page_data)
+        assert (list_data_to_dicts(all_results)) == expected_all_page_data
 
     return inner
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -233,7 +233,7 @@ def test_sync_auto_pagination(
         http_client: Union[SyncHTTPClient, AsyncHTTPClient],
         list_function: Callable[[], WorkOsListResource],
         expected_all_page_data: dict,
-        list_function_params: Optional[Mapping[str, Any]] = {},
+        list_function_params: Optional[Mapping[str, Any]] = None,
     ):
         mock_pagination_request_for_http_client(
             http_client=http_client,
--- a/tests/test_user_management.py
+++ b/tests/test_user_management.py
@@ -1,7 +1,7 @@
 import json
+
 from six.moves.urllib.parse import parse_qsl, urlparse
 import pytest
-import workos
 
 from tests.utils.fixtures.mock_auth_factor_totp import MockAuthFactorTotp
 from tests.utils.fixtures.mock_email_verification import MockEmailVerification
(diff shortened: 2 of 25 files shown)


workos/resources/user_management.py
```diff
@@ -1,232 +1,122 @@
...
-
-    Attributes:
-        OBJECT_FIELDS (list): List of fields a WorkOSInvitation comprises.
-    """
+    access_token: str
+    refresh_token: str
+
+
+class EmailVerification(WorkOSModel):
+    """Representation of a EmailVerification object as returned by WorkOS through User Management features."""
```
PaulAsjes, line 49: Do we need to keep reminding developers that this is through User Management features?

workos/user_management.py
```diff
@@ -1055,189 +981,128 @@ def create_password_reset(
...
+    def reset_password(self, token: str, new_password: str) -> User:
         """Resets user password using token that was sent to the user.
 
         Kwargs:
             token (str): The reset token emailed to the user.
             new_password (str): The new password to be set for the user.
 
         Returns:
-            dict: User response from WorkOS.
+            User: User response from WorkOS.
```
PaulAsjes, line 1006: 👏

workos/user_management.py
```diff
@@ -1312,120 +1144,118 @@ def enroll_auth_factor(
...
-            "params": {
-                "user_id": user_id,
-            },
-            "method": UserManagement.list_auth_factors,
+        # We don't spread params on this dict to make mypy happy
+        list_args: AuthenticationFactorsListFilters = {
+            "limit": limit or DEFAULT_LIST_RESPONSE_LIMIT,
+            "before": before,
+            "after": after,
+            "order": order or "desc",
```
PaulAsjes, line 1202: Previously we changed this to `"order": order` dropping the `or "desc"` part. Should that be the case here?

PaulAsjes (changes requested):
Hard to review a PR of this magnitude, but I found some nits that I think we should address.

workos/user_management.py
```diff
@@ -58,223 +68,349 @@
...
+        last_name: Optional[str] = None,
+        email_verified: Optional[bool] = False,
+    ) -> User: ...
+
+    def update_user(
+        self,
+        user_id: str,
+        first_name: Optional[str] = None,
+        last_name: Optional[str] = None,
+        email_verified: Optional[bool] = False,
```
tribble, line 123: Would this mean that by default we would always pass `False` on update? Would that end up unverifying a user?

workos/user_management.py
```diff
@@ -58,223 +68,349 @@
...
+            params["provider"] = provider
+        if domain_hint is not None:
+            params["domain_hint"] = domain_hint
+        if login_hint is not None:
+            params["login_hint"] = login_hint
+        if state is not None:
+            params["state"] = state
+        if code_challenge:
+            params["code_challenge"] = code_challenge
+            params["code_challenge_method"] = "S256"
```
tribble, line 227: Makes we wonder why we have a param for this if it only has one value

tribble (approved):
One thing I thing needs to be addressed: setting `email_verified=False` by default in user update methods. I'm worried that's going to lead to unexpected behavior.

Otherwise, I don't like that our API JSON models don't match our API reference docs. For now I'm OK going with whatever our ref docs say, but it seems like there should be a follow-up from the teams that own various API surface areas to eliminate the discrepancies.

Thoughts on ☝️ @PaulAsjes?

Slow smile

PaulAsjes:
I agree that `email_verified` shouldn't default to anything. API ref says this field is optional, so we should send `undefined` if it's not explicitly set as presumably the API can handle that.

tests/conftest.py
```diff
@@ -224,3 +223,31 @@ def mock_function(*args, **kwargs):
...
+
+@pytest.fixture
+def test_sync_auto_pagination(
+    mock_pagination_request_for_http_client,
+):
+    def inner(
+        http_client: Union[SyncHTTPClient, AsyncHTTPClient],
+        list_function: Callable[[], WorkOsListResource],
+        expected_all_page_data: dict,
+        list_function_params: Optional[Mapping[str, Any]] = {},
```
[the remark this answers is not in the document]
mattgd: Yes we did, fixing.

tests/test_user_management.py
```diff
@@ -652,261 +459,240 @@ def test_authorization_url_has_expected_query_params_with_code_challenge(self):
...
+        response = self.user_management.authenticate_with_code(**params)
 
-        assert request_args[1].endswith("user_management/authenticate")
-        assert response["user"]["id"] == "user_01H7ZGXFP5C6BBQY6Z7277ZCT0"
-        assert response["impersonator"]["email"] == "admin@foocorp.com"
-        assert response["impersonator"]["reason"] == "Debugging an account issue."
+        assert request_kwargs["url"].endswith("user_management/authenticate")
+        assert response.user.id == "user_01H7ZGXFP5C6BBQY6Z7277ZCT0"
+        assert response.impersonator is not None
+        assert response.impersonator.dict() == {
```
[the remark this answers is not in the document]
mattgd: Hmm, yeah. A bit odd. These are sourced from the dashboard-only, so there's definitely some other magic going on.

workos/resources/mfa.py
```diff
@@ -31,6 +57,33 @@ def to_dict(self):
...
+class AuthenticationFactorTotp(WorkOSModel):
+    """Representation of a MFA Authentication Factor Response as returned by WorkOS through the MFA feature."""
+
+    object: Literal["authentication_factor"]
+    id: str
+    created_at: str
+    updated_at: str
+    type: TotpAuthenticationFactorType
+    totp: Union[TotpFactor, ExtendedTotpFactor, None]
+    user_id: str
```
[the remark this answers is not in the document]
mattgd: Didn't see that, updating.

workos/resources/mfa.py
```diff
@@ -91,6 +144,24 @@ def to_dict(self):
...
 
 
+class AuthenticationChallenge(WorkOSModel):
+    """Representation of a MFA Challenge Response as returned by WorkOS through the MFA feature."""
+
+    object: Literal["authentication_challenge"]
+    id: str
+    created_at: str
+    updated_at: str
+    expires_at: str
```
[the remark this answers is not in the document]
mattgd: Added/changed!

workos/resources/user_management.py
```diff
@@ -1,232 +1,122 @@
...
-            authentication_response_dict["impersonator"] = self.impersonator.to_dict()
+class AuthenticationResponse(WorkOSModel):
+    """Representation of a User and Organization ID response as returned by WorkOS through User Management features."""
 
-        return authentication_response_dict
+    access_token: str
+    impersonator: Optional[Impersonator] = None
+    organization_id: Optional[str] = None
+    refresh_token: str
+    user: User
```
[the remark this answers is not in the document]
mattgd: Yes we do, added!

workos/resources/user_management.py
mattgd: We don't. Removing that copy-pasta.

workos/resources/user_management.py
```diff
@@ -1,232 +1,122 @@
...
+    updated_at: str
+
+
+class Invitation(WorkOSModel):
+    """Representation of an Invitation as returned by WorkOS through User Management features."""
+
+    object: Literal["invitation"]
+    id: str
+    email: str
+    state: str
```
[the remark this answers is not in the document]
mattgd: It should be. Fixed!

workos/user_management.py
mattgd: Yep, that's not correct. Should be `= None` as it's optional and we'll defer to API behavior.

workos/user_management.py
```diff
@@ -283,121 +419,134 @@ def list_users(
...
-        return WorkOSUser.construct_from_response(response).to_dict()
+        return User.model_validate(response)
 
-    def update_user(self, user_id, payload):
+    def update_user(
+        self,
+        user_id: str,
+        first_name: Optional[str] = None,
+        last_name: Optional[str] = None,
+        email_verified: Optional[bool] = False,
```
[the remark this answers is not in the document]
mattgd: Fixed!

workos/user_management.py
mattgd: It should, fixed!

tribble:
@PaulAsjes Longer term I'm more bothered by this:

> Otherwise, I don't like that our API JSON models don't match our API reference docs.

PR #295, opened 2024-07-29
Add typing for MFA

mattgd:
## Description
Add typing for MFA.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/mfa.py
+++ b/workos/mfa.py
@@ -1,64 +1,62 @@
-from typing import Protocol
-from warnings import warn
+from typing import Optional, Protocol
 
 import workos
+from workos.utils.http_client import SyncHTTPClient
 from workos.utils.request import (
-    RequestHelper,
     REQUEST_METHOD_POST,
     REQUEST_METHOD_DELETE,
     REQUEST_METHOD_GET,
+    RequestHelper,
 )
 from workos.utils.validation import MFA_MODULE, validate_settings
 from workos.resources.mfa import (
-    WorkOSAuthenticationFactorSms,
-    WorkOSAuthenticationFactorTotp,
-    WorkOSChallenge,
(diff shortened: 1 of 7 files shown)


workos/utils/request.py
```diff
@@ -48,7 +48,8 @@ def set_base_api_url(self, base_api_url):
     def generate_api_url(self, path):
         return self.base_api_url.format(path)
 
-    def build_parameterized_url(self, url, **params):
+    @classmethod
+    def build_parameterized_url(cls, url, **params):
```
mattgd, line 52: Still on the fence about the proper home for this method. Waiting until we're no longer RequestHelper dependent to move around.

workos/mfa.py
```diff
@@ -78,18 +76,7 @@ def enroll_factor(
...
-        if type not in ["sms", "totp"]:
-            raise ValueError("Type parameter must be either 'sms' or 'totp'")
-
-        if (
-            type == "totp"
-            and totp_issuer is None
-            or type == "totp"
-            and totp_user is None
-        ):
+        if type == "totp" and (totp_issuer is None or totp_user is None):
```
tribble, line 79: @PaulAsjes food for thought: it's hard to enforce these sorts of `if 'totp' then require `issuer` and `user` constraints clearly in a rest API. But we have more options in an SDK.

If we were rewriting this SDK from scratch, would it be better to have separate functions `enroll_sms` and `enroll_totp` with different required params?

tribble (approved):
Ship it.

weirdo gif

workos/mfa.py
mattgd: +1 to this. I thought about having to separate methods, or better yet overloaded methods.

PR #297, opened 2024-07-29
Add typing for Audit Logs APIs

mattgd:
## Description
Add typing for Audit Logs APIs.

## Documentation

```
[ ] Yes
```


workos/resources/audit_logs.py
```diff
@@ -0,0 +1,52 @@
...
+    metadata: NotRequired[dict]
+    name: NotRequired[str]
+    type: str
+
+
+class AuditLogEventTarget(TypedDict):
+    """Describes the entity that was targeted by the event."""
+
+    id: str
+    metadata: NotRequired[dict]
```
tribble, line 33: Oh! We can use `NotRequired` instead of `total=False`.

... but [is it available in 3.8](https://peps.python.org/pep-0655/)?

tribble (approved):
Approved assuming 3.8 supports `NotRequired`

Well Done

workos/resources/audit_logs.py
tribble: Build is green. So that answers that.

mattgd: Yep, but looks like for 3.8 -> 3.10 we have to fetch it from `typing_extensions`.

PR #298, opened 2024-07-29
Add typing for Passwordless APIs

mattgd:
## Description
Add typing for Passwordless APIs.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/passwordless.py
+++ b/workos/passwordless.py
@@ -1,13 +1,21 @@
-from typing import Literal, Protocol
+from typing import Literal, Optional, Protocol
 
 import workos
-from workos.utils.request import RequestHelper, REQUEST_METHOD_POST
+from workos.utils.http_client import SyncHTTPClient
+from workos.utils.request import REQUEST_METHOD_POST
 from workos.utils.validation import PASSWORDLESS_MODULE, validate_settings
-from workos.resources.passwordless import WorkOSPasswordlessSession
+from workos.resources.passwordless import PasswordlessSession, PasswordlessSessionType
 
 
 class PasswordlessModule(Protocol):
-    def create_session(self, session_options: dict) -> dict: ...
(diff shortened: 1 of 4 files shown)


workos/passwordless.py
```diff
@@ -1,61 +1,79 @@
...
+                from WorkOS. Defaults to the default Redirect URI in the
+                WorkOS dashboard.
+            state (str): Optional parameter that the redirect
+                URI received from WorkOS will contain. The state parameter
+                can be used to encode arbitrary information to help
+                restore application state between redirects.
+            type (str): The type of Passwordless Session to
+                create. Currently, the only supported value is 'MagicLink'.
+            expires_in (int): The number of seconds the Passwordless Session should live before expiring.
+                This value must be between 900 (15 minutes) and 86400 (24 hours), inclusive.
```
mattgd, line 55: Changes like these definitely need documenting, but fortunately the developer will be able to take their existing dict and just spread it on to the method call. For example:

4.x usage:
```
session_options = {
    ...
}

create_session(session_options)
```

5.x usage:
```
session_options = {
    ...
}

create_session(**session_options)
```

tribble: ... so this was like the API function we decided should accept a `dict`?

tribble (approved):
yay

workos/passwordless.py
mattgd: I was under the impression we were standardizing on separate parameters like this change. Reason being, once we're on 3.12+, with have [PEP 692](https://peps.python.org/pep-0692/), we can type kwargs with a TypedDict. Also fewer existing methods use this "dict" parameter pattern.

Either way we landed, I'm realizing we probably should've logged that somewhere. Let me know if you think I should swap this back to one dict parameter.

PR #303, opened 2024-07-30
Remove WorkOSBaseResource

mattgd:
## Description
Remove WorkOSBaseResource and refactor test mocks.

## Documentation

```
[ ] Yes
```


tests/test_sso.py
```diff
@@ -15,54 +16,37 @@
...
+            email="demo@workos-magic-link.com",
+            organization_id=None,
+            connection_id="conn_01EMH8WAK20T42N2NBMNBCYHAG",
+            connection_type="MagicLink",
+            idp_id="",
+            first_name=None,
+            last_name=None,
+            groups=None,
+            raw_attributes={},
+        ).dict()
```
mattgd, line 35: Haven't quite landed on the perfect scheme for these mocks. We could add additional flexibility by allowing all parameters to be passed, but provide mock defaults.

tribble: Sounds a lot like a test fixture generation tool, like Ruby's machinist or FactoryGirl. A problem for future us.

One advantage of reusing our existing pydantic models as fixtures is it gives us the ability to include fields in the test response data that are not part of the pydantic model.

workos/resources/list.py
```diff
@@ -24,102 +23,11 @@
 from workos.resources.mfa import AuthenticationFactor
 from workos.resources.organizations import Organization
 from pydantic import BaseModel, Field
-from workos.resources.sso import Connection, ConnectionWithDomains
+from workos.resources.sso import ConnectionWithDomains
 from workos.resources.user_management import Invitation, OrganizationMembership, User
 from workos.resources.workos_model import WorkOSModel
 
 
-class WorkOSListResource(WorkOSBaseResource):
```
tribble, line 32: See Ya

tribble (approved):
Happy Dog

PR #304, opened 2024-07-30
Fix methods missing _id for unique IDs

mattgd:
## Description
Fix methods missing _id for unique IDs.

## Documentation

```
[ ] Yes
```


tribble (approved):
cat waterskiing because why not

PR #307, opened 2024-07-31
Type webhook methods

tribble:
## Description
Add type hints to `verify_event`. Unfortunately, this lead to a bunch of duplication, comments inline.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/resources/webhooks.py
+++ b/workos/resources/webhooks.py
@@ -1,0 +1,283 @@
+from typing import Generic, Literal, Union
+from pydantic import Field
+from typing_extensions import Annotated
+from workos.resources.directory_sync import DirectoryGroup
+from workos.resources.events import EventPayload
+from workos.resources.user_management import OrganizationMembership, User
+from workos.resources.workos_model import WorkOSModel
+from workos.types.directory_sync.directory_user import DirectoryUser
+from workos.types.events.authentication_payload import (
+    AuthenticationEmailVerificationSucceededPayload,
+    AuthenticationMagicAuthFailedPayload,
+    AuthenticationMagicAuthSucceededPayload,
+    AuthenticationMfaSucceededPayload,
+    AuthenticationOauthSucceededPayload,
+    AuthenticationPasswordFailedPayload,
+    AuthenticationPasswordSucceededPayload,
+    AuthenticationSsoSucceededPayload,
+)
+from workos.types.events.connection_payload_with_legacy_fields import (
+    ConnectionPayloadWithLegacyFields,
+)
+from workos.types.events.directory_group_membership_payload import (
+    DirectoryGroupMembershipPayload,
+)
+from workos.types.events.directory_group_with_previous_attributes import (
+    DirectoryGroupWithPreviousAttributes,
+)
+from workos.types.events.directory_payload import DirectoryPayload
+from workos.types.events.directory_payload_with_legacy_fields import (
+    DirectoryPayloadWithLegacyFields,
+)
+from workos.types.events.directory_user_with_previous_attributes import (
+    DirectoryUserWithPreviousAttributes,
+)
+from workos.types.events.organization_domain_verification_failed_payload import (
+    OrganizationDomainVerificationFailedPayload,
+)
+from workos.types.events.session_created_payload import SessionCreatedPayload
+from workos.types.organizations.organization_common import OrganizationCommon
+from workos.types.organizations.organization_domain import OrganizationDomain
+from workos.types.roles.role import Role
+from workos.types.sso.connection import Connection
+from workos.types.user_management.email_verification_common import (
+    EmailVerificationCommon,
+)
+from workos.types.user_management.invitation_common import InvitationCommon
+from workos.types.user_management.magic_auth_common import MagicAuthCommon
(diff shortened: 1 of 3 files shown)


workos/resources/webhooks.py
```diff
@@ -0,0 +1,283 @@
...
+
+class WebhookModel(WorkOSModel, Generic[EventPayload]):
+    """Representation of an Webhook delivered via Webhook.
+    Attributes:
+        OBJECT_FIELDS (list): List of fields an Webhook is comprised of.
+    """
+
+    id: str
+    data: EventPayload
+    created_at: str
```
tribble, line 59: Super-sad panda 😿 

There is a difference between our webhooks and events. Each event contains `object: 'event'`. Our webhooks do not.

A mini future us project should be to add this `object` attributes to webhook bodies so that they're identical and we can remove this sort of duplication.

workos/webhooks.py
```diff
@@ -1,18 +1,32 @@
...
 from workos.utils.request_helper import RequestHelper
 from workos.utils.validation import WEBHOOKS_MODULE, validate_settings
 import hmac
-import json
 import time
-from collections import OrderedDict
 import hashlib
 
+WebhookPayload = Union[bytes, bytearray]
+WebhookTypeAdapter: TypeAdapter[Webhook] = TypeAdapter(Webhook)
```
tribble, line 11: `Webhook` is a discriminated union, not a model. So we need to create a `TypeAdapter` to validate the incoming payload.

workos/webhooks.py
```diff
@@ -34,19 +48,23 @@ def request_helper(self):
...
-
+    def verify_event(
+        self,
+        payload: WebhookPayload,
+        sig_header: str,
+        secret: str,
+        tolerance: Optional[int] = DEFAULT_TOLERANCE,
+    ) -> Webhook:
         Webhooks.verify_header(self, payload, sig_header, secret, tolerance)
-        event = json.loads(payload, object_pairs_hook=OrderedDict)
```
tribble, line 46: Got rid of this and just relying on pydantic's JSON parser (looks like [jiter](https://docs.rs/jiter/latest/jiter/))

tribble:
The webhooks tests still need fixing. The content of the webhook is incorrect, I need to replace it with a valid body and corresponding valid hash.

tribble:
@mattgd Updated tests! Ready for review now.

workos/resources/webhooks.py
```diff
@@ -0,0 +1,283 @@
...
+)
+from workos.types.user_management.invitation_common import InvitationCommon
+from workos.types.user_management.magic_auth_common import MagicAuthCommon
+from workos.types.user_management.password_reset_common import PasswordResetCommon
+
+
+class WebhookModel(WorkOSModel, Generic[EventPayload]):
+    """Representation of an Webhook delivered via Webhook.
+    Attributes:
+        OBJECT_FIELDS (list): List of fields an Webhook is comprised of.
```
mattgd, line 54: This comment can go since `OBJECT_FIELDS` doesn't exist anymore.

workos/webhooks.py
```diff
@@ -34,19 +48,23 @@ def request_helper(self):
...
-    def verify_header(self, event_body, event_signature, secret, tolerance=None):
+        return WebhookTypeAdapter.validate_json(payload)
+
+    def verify_header(
+        self,
+        event_body: WebhookPayload,
+        event_signature: str,
+        secret: str,
+        tolerance: Optional[int] = None,
+    ):
```
mattgd, line 67: Not a task for now, but does needing to wrap `verify_header()` in a `try/except` seem intuitive to you?

mattgd, line 67: We'll want an explicit `-> None` return here.

mattgd (approved):

workos/resources/webhooks.py
tribble: Yeah, I plan to do a comment/docstring sweep across all the resources as a follow-up. Though, I haven't seen any place where these docstrings are surfaced to the developer. Do you know how they're usually used?

workos/webhooks.py
tribble: Good question. It seems like the blessed path is to use `verify_event` (which actually shouldn't be called `verify_event`, it's really a deserialization). It makes sense for that to raise when the header can't be verified.

So what's the use case for using `verify_header` directly? Bypassing deserialization and parsing the webhook into a raw `dict`?

tribble: I'm going to add a linear task to settle this.

tribble: https://linear.app/workos/issue/DSYNC-1775/refactor-webhook-sdk-methods

workos/resources/webhooks.py
mattgd: I see them in my editor, albeit poorly formatted:
image

tribble: Interesting. It seems like they're more useful the the actual SDK methods. Not to much for the resources/models. I have to dig into the packages to actually see the docstrings for our resources.
CleanShot 2024-07-31 at 10 52 18

PR #316, opened 2024-08-05
Fix auth factor response types

tribble:
## Description
WIP

## Documentation

```
[ ] Yes
```


tests/test_mfa.py
```diff
@@ -152,13 +167,13 @@ def test_enroll_factor_totp_success(
...
+        self, mock_get_factor_response_totp, mock_http_client_with_response
     ):
         mock_http_client_with_response(
-            self.http_client, mock_enroll_factor_response_totp, 200
+            self.http_client, mock_get_factor_response_totp, 200
         )
-        response = self.mfa.get_factor(mock_enroll_factor_response_totp["id"])
-        assert response.dict() == mock_enroll_factor_response_totp
+        response = self.mfa.get_factor(mock_get_factor_response_totp["id"])
+        assert response.dict() == mock_get_factor_response_totp
```
tribble, line 176: This test was wrong. The get [auth factor method](https://github.com/workos/workos/blob/57d002f579c73d8a0b266b2356f518ee78bed946/packages/api/src/authentication-factors/authentication-factors.controller.ts#L189) does not return secrets. By [default the serializer does not return secrets](https://github.com/workos/workos/blob/57d002f579c73d8a0b266b2356f518ee78bed946/packages/api/src/authentication-factors/authentication-factor-serializer.ts#L25).

PaulAsjes (approved):

PR #319, opened 2024-08-06
Remove need to call `auto_paging_iter` and use single WorkOsListResource for sync and async

mattgd:
## Description
Remove need to call `auto_paging_iter` and use single WorkOsListResource for sync and async.

With this change, list() methods can be used in two ways:
```
organization_memberships = (
    workos.client.user_management.list_organization_memberships(
        organization_id="org_01HB6C9FZWRPDTVQ3YNBB1Q3JB"
    )
)
```
OR
```
all_organization_memberships = []
for (
    organization_membership
) in workos.client.user_management.list_organization_memberships(
    organization_id="org_01HB6C9FZWRPDTVQ3YNBB1Q3JB",
):
    all_organization_memberships.append(organization_membership)
```

## Documentation

```
[ ] Yes
```

### DIFF
--- a/tests/test_directory_sync.py
+++ b/tests/test_directory_sync.py
@@ -1,5 +1,6 @@
 import pytest
 
+from tests.conftest import test_sync_auto_pagination
 from tests.utils.list_resource import list_data_to_dicts, list_response_of
 from workos.directory_sync import AsyncDirectorySync, DirectorySync
 from workos.utils.http_client import AsyncHTTPClient, SyncHTTPClient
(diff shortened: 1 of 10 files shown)


tests/test_directory_sync.py
```diff
@@ -543,33 +482,10 @@ async def test_directory_user_groups_auto_pagination(
...
         assert (
             list_data_to_dicts(all_groups)
         ) == mock_directory_groups_multiple_data_pages
-
-    async def test_auto_pagination_honors_limit(
-        self,
-        mock_directories_multiple_data_pages,
-        mock_pagination_request_for_http_client,
-    ):
-        # TODO: This does not actually test anything about the limit.
```
tribble, line 559: We lost this TODO. I still don't think we're testing inputs. Though, how *should* limit work with iteration? Is it a limit on the total number of returned records?

tests/test_organizations.py
```diff
@@ -181,7 +181,7 @@ def test_list_organizations_auto_pagination_for_single_page(
         all_organizations = []
         organizations = self.organizations.list_organizations()
 
-        for org in organizations.auto_paging_iter():
+        for org in organizations:
```
tribble, line 184: This is lots better

workos/resources/list.py
```diff
@@ -135,12 +131,7 @@ def auto_paging_iter(self) -> Iterator[ListableResource]:
             yield self.data[index]
             index += 1
 
-
-class AsyncWorkOsListResource(
-    BaseWorkOsListResource,
-    Generic[ListableResource, ListAndFilterParams, ListMetadataType],
-):
-    async def auto_paging_iter(self) -> AsyncIterator[ListableResource]:
+    async def __aiter__(self) -> AsyncIterator[ListableResource]:
```
tribble, line 134: Pronounced... ate-er?

tribble (approved):
Wondering about how `limit` actually plays with iteration. But that can be a separate follow-up.

Great Job

tests/test_directory_sync.py
mattgd: This is a great question, I'm not sure. Worth us testing out, but also maybe playing with it in an example app/looking for prior art.

PR #320, opened 2024-08-06
Change unlinked directory state to inactive and convert in validator

tribble:
## Description
We made this conversion for `active` but forgot to do so for `inactive`

Resolves DSYNC-1785

## Documentation

```
[ ] Yes
```

### DIFF
--- a/tests/types/test_directory_state.py
+++ b/tests/types/test_directory_state.py
@@ -21,25 +21,14 @@ class TestDirectoryState:
         assert directory_state_type_adapter.validate_python("unlinked") == "inactive"
         assert directory_state_type_adapter.validate_python("inactive") == "inactive"
 
+    @pytest.mark.parametrize(
+        "type_value", ["foo", None, 5, {"definitely": "not a state"}]
+    )
     def test_invalid_values_returns_validation_error(
-        self, directory_state_type_adapter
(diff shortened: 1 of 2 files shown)


tests/types/test_directory_state.py
```diff
@@ -0,0 +1,45 @@
...
+
+        try:
+            directory_state_type_adapter.validate_python(5)
+        except ValidationError as e:
+            assert e.errors()[0]["type"] == "literal_error"
+
+        try:
+            directory_state_type_adapter.validate_python({"definitely": "not a state"})
+        except ValidationError as e:
+            assert e.errors()[0]["type"] == "literal_error"
```
mattgd, line 45: Could be a good place for `@pytest.mark.parameterize` decorator. Something like:
```
@pytest.mark.parametrize("type_value", [("foo"), (None), (5), ({"definitely": "not a state"})])
```

mattgd (approved):

tests/types/test_directory_state.py
tribble: Oooh. Never used this. Let me look it up and make the change.

tribble: Ooh. Cute. More strings for referencing types or args. But I'll go with it.

PR #321, opened 2024-08-07
api: Add async support for UM

mattgd:
## Description
Add async support for UM.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/user_management.py
+++ b/workos/user_management.py
@@ -34,7 +34,7 @@ from workos.types.user_management.authenticate_with_common import (
     AuthenticateWithTotpParameters,
 )
 from workos.typing.sync_or_async import SyncOrAsync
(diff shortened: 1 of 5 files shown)


workos/user_management.py
```diff
@@ -1385,3 +1381,976 @@ def revoke_invitation(self, invitation_id: str) -> Invitation:
         )
 
         return Invitation.model_validate(response)
+
+
+class AsyncUserManagement(UserManagementModule):
```
mattgd, line 1386: I really dislike the duplication here, on all levels: method definitions, business logic, doc strings. I looked into how we could remove the duplication, but a lot of the potentially solutions I found were fairly complex. This is something we should look into more.

tribble (approved):
:shipit:

PR #327, opened 2024-08-08
Allow multiple WorkOS clients to be instantiated with different keys

mattgd:
## Description
Allow multiple WorkOS clients to be instantiated with different keys.

## Documentation

```
[ ] Yes
```


tests/test_client.py
```diff
@@ -1,219 +1,109 @@
...
-        assert bool(client.user_management)
-
-    def test_initialize_sso_missing_api_key(self, set_client_id):
-        with pytest.raises(ConfigurationException) as ex:
-            client.sso
-
-        message = str(ex)
-
-        assert "api_key" in message
-        assert "client_id" not in message
```
mattgd, line 53: Removed a bunch of "module"-specific tests like this and replaced with a few simple tests for apiKey/clientId success/error cases.

workos/async_client.py
```diff
@@ -17,20 +20,26 @@ class AsyncClient(BaseClient):
...
-    _audit_logs: AuditLogsModule
-    _directory_sync: AsyncDirectorySync
-    _events: AsyncEvents
-    _mfa: MFAModule
-    _organizations: OrganizationsModule
-    _passwordless: PasswordlessModule
-    _portal: PortalModule
-    _sso: AsyncSSO
-    _user_management: AsyncUserManagement
-    _webhooks: WebhooksModule
```
mattgd, line 29: I thought these were necessary for mypy, but apparently not?

workos/__init__.py
```diff
@@ -1,16 +1,5 @@
...
-base_api_url = os.getenv("WORKOS_BASE_URL", "https://api.workos.com/")
-request_timeout = int(os.getenv("WORKOS_REQUEST_TIMEOUT", "25"))
-
-
-client = SyncClient(base_url=base_api_url, version=__version__, timeout=request_timeout)
-async_client = AsyncClient(
-    base_url=base_api_url, version=__version__, timeout=request_timeout
-)
+WorkOSClient = SyncClient
+AsyncWorkOSClient = AsyncClient
```
mattgd, line 5: We could also just rename the clients and re-export. Thoughts?

workos/utils/http_client.py
```diff
@@ -168,7 +174,7 @@ async def request(
         params: ParamsType = None,
         json: JsonType = None,
         headers: HeadersType = None,
-        token: Optional[str] = None,
+        exclude_default_auth_headers: bool = False,
```
mattgd, line 177: New param used for UM auth methods/a few other methods where we don't need the regular Auth header.

workos/_base_client.py
```diff
@@ -13,37 +17,98 @@
 from workos.webhooks import WebhooksModule
 
 
-class BaseClient(Protocol):
+HTTPClientType = TypeVar("HTTPClientType", bound=HTTPClient)
+
+
+class BaseClient(Generic[HTTPClientType]):
```
mattgd, line 23: Did a bit of a rework here to pass in the HTTP client type, partly because Python versions <3.11 were having issues resolving the _http_client attribute properly.

tests/test_client.py
```diff
@@ -1,219 +1,109 @@
...
-        message = str(ex)
-
-        assert "api_key" in message
-
-    def test_initialize_events_missing_api_key(self):
-        with pytest.raises(ConfigurationException) as ex:
-            client.events
-
-        message = str(ex)
+    def test_client_without_api_key(self):
```
tribble, line 13: If no `api_key` is provided, we throw an error. Do we need this fixture?

tests/test_client.py
```diff
@@ -1,219 +1,109 @@
...
-        assert "client_id" in message
+        assert (
+            "WorkOS API key must be provided when instantiating the client or via the WORKOS_API_KEY environment variable."
+            == str(error.value)
+        )
 
-    def test_initialize_user_management_missing_api_key(self, set_client_id):
-        with pytest.raises(ConfigurationException) as ex:
-            client.user_management
+    def test_client_without_client_id(self):
```
tribble, line 22: Talking through why we need this fixture at all: the SDK has no notion of which API methods need a client_id, so this is essentially emulating an error response from the API?

workos/__init__.py
tribble: I *think* you can also do something like...
`from workos.client import SyncClient as WorkOSClient`

workos/_base_client.py
```diff
@@ -13,37 +17,98 @@
...
+        api_key = api_key or os.getenv("WORKOS_API_KEY")
+        if api_key is None:
+            raise ValueError(
+                "WorkOS API key must be provided when instantiating the client or via the WORKOS_API_KEY environment variable."
+            )
+
+        self._api_key = api_key
+
+        client_id = client_id or os.getenv("WORKOS_CLIENT_ID")
+        if client_id is None:
```
tribble, line 50: I don't think all of our APIs require a `client_id`, should we always require it? The Node SDK does not throw for a missing `client_id`

workos/async_client.py
tribble: Less boilerplate good.

workos/audit_logs.py
```diff
@@ -65,11 +62,7 @@ def create_event(
             headers["idempotency-key"] = idempotency_key
 
         self._http_client.request(
-            EVENTS_PATH,
-            method=REQUEST_METHOD_POST,
-            json=json,
-            headers=headers,
-            token=workos.api_key,
```
tribble, line 72: It was silly that we passed this every time

workos/utils/_base_http_client.py
```diff
@@ -196,6 +215,19 @@ def base_url(self, url: str) -> None:
...
 
+    @property
+    def client_id(self) -> str:
+        return self._client_id
+
+    @property
+    def auth_headers(self) -> Dict[str, str]:
+        return self.auth_header_from_token(self._api_key)
+
+    def auth_header_from_token(self, token: str) -> Dict[str, str]:
```
tribble, line 226: `Mapping`? :)

workos/utils/http_client.py
tribble: Ah, yeah. UM complicates this.

tribble (approved):
How do you want to handle the `client_id` requirement?

Ahhh

workos/utils/_base_http_client.py
mattgd: :D, yes

tests/test_client.py
mattgd: This is testing exactly that: throws an error when API key is missing. The test name is a bit confusing though. "Client" in this case is the API client.

tests/test_client.py
mattgd: Yep, that's correct. If we decide we shouldn't required a client ID for every instantiation, we'll want to rework.

PR #328, opened 2024-08-09
[FGA-73] Add BatchCheck and Query methods to FGA module

atainter:
## Description
Adds CheckBatch and Query methods to FGA module. 

Other enhancements:
- Adds method documentation in the FGA module.
- Fixed an issue with incorrect return types.
- Require named params for readability and backwards compatibility


tests/test_fga.py
```diff
@@ -463,3 +467,79 @@ def test_check_with_debug_info(
...
+                        "resource_type": "role",
+                        "resource_id": "manager",
+                        "relation": "member",
+                        "subject": {"resource_type": "user", "resource_id": "tom"},
+                    },
+                    "is_implicit": True,
+                },
+            ],
+            "list_metadata": {},
+        }
```
mattgd, line 535: No change required, but we added a `list_response_of` helper for these types of responses.

workos/fga.py
```diff
@@ -34,16 +39,21 @@
...
+]
+
 
 class FGAModule(Protocol):
-    def get_resource(self, resource_type: str, resource_id: str) -> Resource: ...
+    def get_resource(self, *, resource_type: str, resource_id: str) -> Resource: ...
 
     def list_resources(
         self,
+        *,
```
mattgd, line 56: Thanks for restricting these. We'll see about adding a linter rule.

mattgd (approved):

PR #332, opened 2024-08-12
Small typing fixes

tribble:
## Description
Note: This is stacked on #331

OK, the branch name is a little LOL

This started as an attempt to fix some of the stricter validations from pylance, like shadowing the `_http_client` type. This led to a bit of a rabbit hole, but here was the outcome:

- Extracted a client configuration protocol, the base modules do not need access to the client, just the base URL and client ID
- The initialization of the base client, which would then initialize the HTTP based on the Generic type in was a bit confusing to me, so instead I directory initialize the HTTP client in the Sync or Async client. This eliminated the need for a generic in the base client entirely

I'm not married to the bigger changes I made around removing the generics for the `BaseClient`, but I feel like this is a little more direct and easier to understand.

## Documentation

```
[ ] Yes
```


workos/_base_client.py
```diff
@@ -117,3 +113,15 @@ def user_management(self) -> UserManagementModule: ...
     @property
     @abstractmethod
     def webhooks(self) -> WebhooksModule: ...
+
+    @property
+    def base_url(self) -> str:
+        return self._base_url
```
mattgd, line 119: In base HTTP client we have a setter that enforces a trailing slash. Would be good to make sure these values are the same and perhaps enforce the trailing slash here instead:
https://github.com/workos/workos-python/blob/6d7ab8c39dce654c1a3fab3642df2533718da518/workos/utils/_base_http_client.py#L209-L216

mattgd (commented):
I see this is still in draft, but left one comment.

mattgd (approved):

PR #334, opened 2024-08-13
Add async support for organizations

mattgd:
## Description
Add async support for organizations. This is the last module I felt like async support could absolutely be useful, in tandem with other modules.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/tests/test_organizations.py
+++ b/tests/test_organizations.py
@@ -1,16 +1,11 @@
 import datetime
 import pytest
 from tests.utils.list_resource import list_data_to_dicts, list_response_of
-from workos.organizations import Organizations
+from workos.organizations import AsyncOrganizations, Organizations
 from tests.utils.fixtures.mock_organization import MockOrganization
 
 
-class TestOrganizations(object):
-    @pytest.fixture(autouse=True)
-    def setup(self, sync_http_client_for_test):
-        self.http_client = sync_http_client_for_test
-        self.organizations = Organizations(http_client=self.http_client)
-
+class OrganizationFixtures:
     @pytest.fixture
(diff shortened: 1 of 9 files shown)


workos/types/webhooks/webhook.py
```diff
@@ -1,11 +1,9 @@
-from typing import Generic, Literal, Union
```
mattgd, line 1: Cleaning up unused imports.

tests/test_organizations.py
```diff
@@ -200,3 +202,142 @@ def test_list_organizations_auto_pagination_for_multiple_pages(
             list_function=self.organizations.list_organizations,
             expected_all_page_data=mock_organizations_multiple_data_pages["data"],
         )
+
+
+@pytest.mark.asyncio
+class TestAsyncOrganizations(OrganizationFixtures):
```
tribble, line 208: This repetition is so grrrrr.

workos/organizations.py
```diff
@@ -75,29 +78,39 @@ def create_organization(
         name: str,
         domain_data: Optional[Sequence[DomainDataInput]] = None,
         idempotency_key: Optional[str] = None,
-    ) -> Organization: ...
+    ) -> SyncOrAsync[Organization]:
+        """Create an organization
```
tribble, line 82: Odd. I wonder why this is in the diff Is it actually missing?

tribble (approved):
Great Job

tests/test_organizations.py
mattgd: I know 😢 We need some more thought to strike the right balance between little repetition and and readability.

PR #337, opened 2024-08-14
Rename FGA types

tribble:
## Description
Rename `Resource` because we also use that term for endpoints, like `ListResource`.

Upon further though, I'm now on the fence with this one. The `Resource` is nested in the FGA types. And I'm not loving any of the renames.

`AuthorizationResource` - Is that what this is? Is it actually an `AuthorizableResource`? So far I've gone with `AuthorizationResource`.
`FGAResource` - Meh, I don't love acronyms.
`WarrantResource` - Feels like this is shipping our org structure and just prefixing with `Warrant`.

Stacked on #336  if we want to merge this. I'd love the reviewers thoughts.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/fga.py
+++ b/workos/fga.py
@@ -2,8 +2,8 @@ from typing import Any, Mapping, Optional, Protocol, Sequence
 from workos.types.fga import (
     CheckOperation,
     CheckResponse,
-    Resource,
-    ResourceType,
+    AuthorizationResource,
+    AuthorizationResourceType,
     Warrant,
     WarrantCheckInput,
     WarrantWrite,
@@ -12,9 +12,9 @@ from workos.types.fga import (
     WarrantQueryResult,
 )
 from workos.types.fga.list_filters import (
-    ResourceListFilters,
+    AuthorizationResourceListFilters,
     WarrantListFilters,
-    QueryListFilters,
+    WarrantQueryListFilters,
 )
 from workos.types.list_resource import (
     ListArgs,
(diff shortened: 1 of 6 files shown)


workos/fga.py
```diff
@@ -358,10 +364,12 @@ def list_resources(
             params=list_params,
         )
 
-        return WorkOSListResource[Resource, ResourceListFilters, ListMetadata](
+        return WorkOSListResource[
+            AuthorizationResource, AuthorizationResourceListFilters, ListMetadata
+        ](
```
mattgd, line 369:
```suggestion
        return AuthorizationResourceListResource(
```

workos/fga.py
```diff
@@ -626,7 +634,9 @@ def query(
         # A workaround to add warrant_token to the list_args for the ListResource iterator
         list_params["warrant_token"] = warrant_token
 
-        return WorkOSListResource[WarrantQueryResult, QueryListFilters, ListMetadata](
+        return WorkOSListResource[
+            WarrantQueryResult, WarrantQueryListFilters, ListMetadata
+        ](
```
mattgd, line 639:
```suggestion
        return WarrantQueryListResource(
```

mattgd (approved):
Thanks for updating! Two small suggestions.

workos/fga.py
tribble: Nice catch. Will fix all of these.

atainter (approved):

PR #338, opened 2024-08-14
Fix list_organization_memberships statuses parameter when auto paging

mattgd:
## Description
Fix list_organization_memberships statuses parameter when auto paging.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/user_management.py
+++ b/workos/user_management.py
@@ -947,7 +947,7 @@ class UserManagement(UserManagementModule):
         params: OrganizationMembershipsListFilters = {
             "user_id": user_id,
             "organization_id": organization_id,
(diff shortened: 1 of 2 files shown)


workos/user_management.py
```diff
@@ -947,7 +947,7 @@ def list_organization_memberships(
         params: OrganizationMembershipsListFilters = {
             "user_id": user_id,
             "organization_id": organization_id,
-            "statuses": ",".join(statuses) if statuses else None,
+            "statuses": statuses,
```
mattgd, line 950: This was unnecessary. It's parsed fine by the API as a sequence/list.

PavanKulkarni (approved):

PR #344, opened 2024-08-16
Add missing LiteralOrUntyped wrappers

mattgd:
## Description
Make OrganizationMembership.status LiteralOrUntyped.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/types/user_management/organization_membership.py
+++ b/workos/types/user_management/organization_membership.py
@@ -2,6 +2,7 @@ from typing import Literal
 from typing_extensions import TypedDict
 
 from workos.types.workos_model import WorkOSModel
+from workos.typing.literals import LiteralOrUntyped
 
 OrganizationMembershipStatus = Literal["active", "inactive", "pending"]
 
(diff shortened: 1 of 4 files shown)


workos/types/user_management/organization_membership.py
```diff
@@ -18,6 +19,6 @@ class OrganizationMembership(WorkOSModel):
     user_id: str
     organization_id: str
     role: OrganizationMembershipRole
-    status: OrganizationMembershipStatus
+    status: LiteralOrUntyped[OrganizationMembershipStatus]
```
tribble, line 22: Nice catch.

I wonder if there are other literals like this? In FGA perhaps?

tribble (approved):
Catch

workos/types/user_management/organization_membership.py
tribble: Oh. Yup. Here: https://github.com/workos/workos-python/blob/f1dbfca2e9e9f2cb492fed7748e195dc65608bdd/workos/fga.py#L204

tribble: I only see literals as inputs in FGA actually. So that's fine.

tribble: Oh, but there is also this one in MFA: https://github.com/workos/workos-python/blob/f1dbfca2e9e9f2cb492fed7748e195dc65608bdd/workos/types/mfa/authentication_factor.py#L41

mattgd: Didn't see them at first, but found two others and fixed.

Ndiayecheikh16 (commented):
Leave

PR #351, opened 2024-08-20
Reduce sync/async test code duplication

mattgd:
## Description
PoC for reducing sync/async test code duplication.

## Documentation

```
[ ] Yes
```


tests/test_events.py
```diff
@@ -29,41 +32,7 @@ def test_list_events(self, mock_events, capture_and_mock_http_client_request):
...
-
-    async def test_list_events(self, mock_events, capture_and_mock_http_client_request):
-        request_kwargs = capture_and_mock_http_client_request(
-            http_client=self.http_client,
-            status_code=200,
-            response_dict=mock_events,
-        )
-
-        events = await self.events.list_events(events=["dsync.activated"])
+        events = run_sync(self.events.list_events(events=["dsync.activated"]))
```
mattgd, line 35: Sure, we're not running the code in an async context when under test, but the primary purpose of the test is to ensure the "business" logic of sync/async versions of methods works in the same way.

tests/test_events.py
```diff
@@ -1,14 +1,17 @@
...
 
 
+@pytest.mark.sync_and_async({"sync": Events, "async": AsyncEvents})
 class TestEvents(object):
     @pytest.fixture(autouse=True)
-    def setup(self, sync_http_client_for_test):
-        self.http_client = sync_http_client_for_test
-        self.events = Events(http_client=self.http_client)
+    def setup(self, http_client, module_class):
+        print(http_client)
```
mattgd, line 12: Oops, removing.

tests/utils/syncify.py
```diff
@@ -0,0 +1,19 @@
...
+        return asyncio.run(coro)
+    else:
+        return loop.run_until_complete(coro)
+
+
+def run_sync(obj: Union[Coroutine[Any, Any, Any], Any]) -> Any:
+    if isinstance(obj, CoroutineType):
+        return _call(obj)
+
+    return obj
```
mattgd, line 19: This typing 😢, I might be able to do better?

tests/test_events.py
```diff
@@ -1,14 +1,16 @@
 import pytest
 
+from tests.utils.syncify import run_sync
 from tests.utils.fixtures.mock_event import MockEvent
 from workos.events import AsyncEvents, Events
 
 
+@pytest.mark.sync_and_async({"sync": Events, "async": AsyncEvents})
```
tribble, line 8: I guess this doesn't take kwargs?

tests/conftest.py
```diff
@@ -12,26 +12,70 @@
...
         base_url="https://api.workos.test/",
         client_id="client_b27needthisforssotemxo",
         version="test",
+    ),
+}
+
+
+def pytest_configure(config) -> None:
+    config.addinivalue_line(
+        "markers", "sync_and_async(): mark test to run both sync and async versions"
```
tribble, line 33: In the end, we're not actually running async tests, we're just testing multiple clients/modules, right? How different is this from just supplying a list or tuple of modules rather than naming?

tests/utils/syncify.py
```diff
@@ -0,0 +1,19 @@
...
+def _call(coro: CoroutineType) -> Any:
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:
+        return asyncio.run(coro)
+    else:
+        return loop.run_until_complete(coro)
+
+
+def run_sync(obj: Union[Coroutine[Any, Any, Any], Any]) -> Any:
```
tribble, line 15: What are the types that actually appear here?

tests/test_events.py
mattgd: I'm pretty sure it does actually, I should just do that. Updating!

tests/utils/syncify.py
mattgd: It's either the return type of any of our API methods, or a Coroutine with the return type of any of our API methods.

tribble: Pretty broad. In fact, does that union basically collapse down to `Any`? :)

tests/utils/syncify.py
tribble: I feel like the only way this get better is with a generic parameter, right? The return type should be a reflection of the incoming function return type. That might actually clean up the input param types as well.

tests/conftest.py
mattgd: Without some sort of sync and async parent class, I'm not sure there's a great way to do this.

tribble: Parent class, or `Protocol`? What's the need?

tests/utils/syncify.py
mattgd: Haha, well I guess so 😭

tests/conftest.py
mattgd: We need to know if the `http_client` on the module should be `sync` or `async`. The `Protocol` allows either.

tests/utils/syncify.py
mattgd: Yep, let me see if I can drill this down a bit.

tests/conftest.py
tribble: ... could the http client be passed in with the annotation string? Is that crazy?

tests/utils/syncify.py
tribble: Happy to pair on this if it helps. I know introducing generics can sometimes descend into madness

mattgd: What do you think of the updated version? If you think we can do better, down to pair. I'd love for `run_sync`'s signature to be
```
run_sync(obj: Union[Coroutine[Any, Any, ReturnType], ReturnType]) -> ReturnType:
```
but not sure that's feasible.

tribble: I think that would be great.

tribble (approved):
cute doge

PR #353, opened 2024-08-22
Use syncify for remaining async test suites

mattgd:
## Description
Use syncify for remaining async test suites.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -118,20 +118,8 @@ def sync_http_client_for_test():
 
 
 @pytest.fixture
-def async_http_client_for_test():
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -117,24 +117,12 @@ def sync_http_client_for_test():
     return http_client
 
 
-@pytest.fixture
-def async_http_client_for_test():
-    _, _, http_client = _get_test_client_setup("AsyncHTTPClient")
-    return http_client
```
mattgd, line 123: Not used. Sync versions are used for sync-only modules, but could eventually be removed.

tribble (approved):
a cat wearing bunny ears

PR #360, opened 2024-09-09
Fix authentication error payload

PaulAsjes:
## Description
Exceptions were only returning specific hardcoded values, which isn't accurate any more since an error can have a plethora of parameters in it. 

This fixes it by dumping the contents of the API response into the error object, giving the user all the information they need to fix the error.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/tests/test_sync_http_client.py
+++ b/tests/test_sync_http_client.py
@@ -204,7 +204,7 @@ class TestSyncHTTPClient(object):
             # This'll fail for sure here but... just using the nice error that'd come up
             assert ex.__class__ == expected_exception
 
-    def test_bad_request_exceptions_include_expected_request_data(self):
+    def test_bad_request_exceptions_include_request_data(self):
         request_id = "request-123"
         error = "example_error"
         error_description = "Example error description"
@@ -212,7 +212,7 @@ class TestSyncHTTPClient(object):
         self.http_client._client.request = MagicMock(
             return_value=httpx.Response(
                 status_code=400,
-                json={"error": error, "error_description": error_description},
+                json={"error": error, "error_description": error_description, "foo": "bar"},
                 headers={"X-Request-ID": request_id},
             ),
         )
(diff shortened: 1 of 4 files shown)


tests/test_sync_http_client.py
```diff
@@ -222,29 +222,11 @@ def test_bad_request_exceptions_include_expected_request_data(self):
         except BadRequestException as ex:
             assert (
                 str(ex)
-                == "(message=No message, request_id=request-123, error=example_error, error_description=Example error description)"
+                == "(message=No message, error=example_error, error_description=Example error description, foo=bar, request_id=request-123)"
             )
         except Exception as ex:
             assert ex.__class__ == BadRequestException
 
-    def test_bad_request_exceptions_exclude_expected_request_data(self):
```
PaulAsjes, line 230: What we want is for everything in the return value to be added to exceptions, this test does the opposite.

workos/exceptions.py
```diff
@@ -8,70 +8,35 @@ class BaseRequestException(Exception):
...
-            self.error_description = None
-            self.pending_authentication_token = None
-
-        headers = response.headers
-        self.request_id = headers.get("X-Request-ID")
-
-    def __str__(self) -> str:
-        message = self.message or "No message"
-        exception = "(message=%s" % message
+        self.message = self.extractFromJson("message", "No message")
```
gcarvelli, line 18: This is 100x nicer than `except ValueError`, good stuff

workos/exceptions.py
```diff
@@ -8,70 +8,35 @@ class BaseRequestException(Exception):
...
+        self.error = self.extractFromJson("error", None)
+        self.errors = self.extractFromJson("errors", None)
+        self.code = self.extractFromJson("code", None)
+        self.error_description = self.extractFromJson("error_description", None)
```
gcarvelli, line 22: `error` and `error_description` previously defaulted to `"Unknown"` so this is a small change in behavior ([line](https://github.com/workos/workos-python/pull/360/files#diff-fad3453bb4c7313cf5431a3e0e2939488c284ef8cb50b2f250ed77d4fd8035f2L108))

gcarvelli (approved):

workos/exceptions.py
PaulAsjes: Oh good catch!

workos/exceptions.py
PaulAsjes: Thanks! Also lol snake case

azuline:
thank you!

PR #369, opened 2024-09-27
Fix dsync.activated event parsing for Event API when domains present

mattgd:
## Description
Fix incorrect `dsync.activated` domains type for Events API, causing a parsing error when listing `dsync.activated` events with domains.

This PR leaves the webhooks event payload as a dictionary. This will be changed in the next major version.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/tests/test_webhooks.py
+++ b/tests/test_webhooks.py
@@ -1,5 +1,6 @@
 import json
 import pytest
+from workos.typing.webhooks import WebhookTypeAdapter
 from workos.webhooks import Webhooks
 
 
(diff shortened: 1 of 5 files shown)


tests/test_webhooks.py
```diff
@@ -122,3 +123,32 @@ def test_unrecognized_webhook_type_returns_untyped_webhook(
         )
         assert type(result).__name__ == "UntypedWebhook"
         assert result.dict() == json.loads(mock_unknown_webhook_body)
+
+    def test_validate_dsync_activated_event(self):
```
mattgd, line 127: It would be nice to generalize and parameterize a test like this, though we'd need all the input data.

workos/types/events/directory_payload_with_legacy_fields.py
```diff
@@ -5,7 +5,7 @@
 
 class MinimalOrganizationDomain(WorkOSModel):
     id: str
-    organization_id: str
+    domain: str
```
mattgd, line 8: This is technically a breaking change.

tribble: Is that because we are going to start returning correct objects instead of dicts?

tribble: A key that, right now, is always `None`.

What I'm thinking about is this: since this type is wrong, it's always parsing the Webhook incorrectly, returning a `dict`.

The behavior is worse if they try to retrieve Events. Those don't have a `dict` fallback mechanism, because the client needs to explicitly send the list of event types they want. I think retrieving this type of Event flat-out error because it fails to parse.

If I'm correct about this above, this seems like a more urgent bug fix to me.

PaulAsjes: Talked about this in DMs but bringing the convo here, I think this is unfortunately going to require a major version release. Anyone who is currently using the SDK for webhooks and not listing events is going to have their integration broken if we push this out as a minor or patch.

PaulAsjes (approved):

PR #383, opened 2024-11-20
Add prompt option to creating authorization urls

salbito-workos:
## Description

This changset adds the prompt option to generating oauth authorization urls.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/user_management.py
+++ b/workos/user_management.py
@@ -350,7 +350,9 @@ class UserManagementModule(Protocol):
             state (str): An encoded string passed to WorkOS that'd be preserved through the authentication workflow, passed
                 back as a query parameter. (Optional)
             code_challenge (str): Code challenge is derived from the code verifier used for the PKCE flow. (Optional)
-            prompt (str): The prompt parameter can be used to specify whether the user should be prompted for credentials or not.
+            prompt (str): Used to specify whether the upstream provider should prompt the user for credentials or other
+                consent. Valid values depend on the provider. Currently only applies to provider values of 'GoogleOAuth',
+                'MicrosoftOAuth', or 'GitHubOAuth'. (Optional)
 
         Returns:
             str: URL to redirect a User to to begin the OAuth workflow with WorkOS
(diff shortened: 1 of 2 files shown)


workos/user_management.py
```diff
@@ -349,6 +350,9 @@ def get_authorization_url(
             state (str): An encoded string passed to WorkOS that'd be preserved through the authentication workflow, passed
                 back as a query parameter. (Optional)
             code_challenge (str): Code challenge is derived from the code verifier used for the PKCE flow. (Optional)
+            prompt (str): Used to specify whether the upstream provider should prompt the user for credentials or other
+                consent. Valid values depend on the provider. Currently only applies to provider values of 'GoogleOAuth',
+                'MicrosoftOAuth', or 'GitHubOAuth'. (Optional)
```
mattgd, line 355: Looks like [Google OAuth](https://developers.google.com/identity/openid-connect/openid-connect#authenticationuriparameters), [Microsoft OAuth](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#request-an-authorization-code), and [GitHub OAuth](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps) all have a pretty limited set of values. Could be nice to maintain a set of allowed values with a literal. Not required, though if we want to do it, would be good to do it now, otherwise it'd be a breaking change later.

mattgd (approved):
Left one optional suggestion, but looks good.

workos/user_management.py
mthadley: Since the values are meaningful to the downstream providers (not us), are not consistent across them (though somewhat overlap), and if any providers changed later would require us to update the SDK before new values could be passed, I'm leaning towards skipping validation in order to stay decoupled.

Not a bad idea! Just not sure it's worth it in this particular case.

mattgd: Yep, that makes sense!

PR #389, opened 2024-11-26
Propagate ConflictExceptions properly

mattgd:
## Description
Propagate ConflictExceptions properly and don't mask them as BadRequestExceptions.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/tests/test_sync_http_client.py
+++ b/tests/test_sync_http_client.py
@@ -201,8 +201,6 @@ class TestSyncHTTPClient(object):
         except expected_exception as ex:  # type: ignore
             assert ex.message == response_message
             assert ex.request_id == request_id
-        except Exception as ex:
-            # This'll fail for sure here but... just using the nice error that'd come up
             assert ex.__class__ == expected_exception
 
     def test_bad_request_exceptions_include_request_data(self):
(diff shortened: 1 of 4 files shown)


tests/test_sync_http_client.py
```diff
@@ -251,6 +252,23 @@ def test_request_bad_body_raises_expected_exception_with_request_data(self):
...
+                headers={"X-Request-ID": request_id},
+            ),
+        )
+
+        try:
+            self.http_client.request("bad_place")
+        except ConflictException as ex:
+            assert str(ex) == "(message=No message, request_id=request-123)"
+        except Exception as ex:
+            assert ex.__class__ == ConflictException
```
gcarvelli, line 270: Do you know why we do this multiple-`except` blocks in this test file?

IIRC Python will only call the first matching `except` block so this `except Exception` block will never be entered.

gcarvelli (approved):

tests/test_sync_http_client.py
mattgd: That's a great question. After changing the test it appears there's no reason really. Let me update these.

mattgd: Also confirming one of these tests failing doesn't look any different.

PR #392, opened 2024-12-12
Add GET /organization/:orgId/roles support

mattgd:
## Description
Add GET /organization/:orgId/roles support.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/organizations.py
+++ b/workos/organizations.py
@@ -2,7 +2,7 @@ from typing import Optional, Protocol, Sequence
 
 from workos.types.organizations.domain_data_input import DomainDataInput
 from workos.types.organizations.list_filters import OrganizationListFilters
-from workos.types.roles.role import OrganizationRole, RolesList
+from workos.types.roles.role import RoleList
 from workos.typing.sync_or_async import SyncOrAsync
 from workos.utils.http_client import AsyncHTTPClient, SyncHTTPClient
 from workos.utils.pagination_order import PaginationOrder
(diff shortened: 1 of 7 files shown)


workos/types/roles/role.py
```diff
@@ -1,8 +1,28 @@
...
 
-class Role(WorkOSModel):
+
+class RoleCommon(WorkOSModel):
     object: Literal["role"]
     slug: str
+
+
+# TODO: This is used for events/webhooks only. Rename to EventRole or something similar.
+class Role(RoleCommon):
```
mattgd, line 13: This is a bit awkward. Wish I could rename, but that'd be breaking. Really we should unify the type returned by events and the roles API.

workos/organizations.py
```diff
@@ -223,6 +224,14 @@ def delete_organization(self, organization_id: str) -> None:
             method=REQUEST_METHOD_DELETE,
         )
 
+    def list_organization_roles(self, organization_id: str) -> RolesList:
```
mattgd, line 227: How do we feel about this being housed in the organizations module? It's a subresource in the API, but should be separated as such? Alternatively we could introduce the idea of a sub-module.

mattgd: See discussion in Slack.

workos/organizations.py
```diff
@@ -223,6 +224,14 @@ def delete_organization(self, organization_id: str) -> None:
...
+        response = self._http_client.request(
+            f"organizations/{organization_id}/roles",
+            method=REQUEST_METHOD_GET,
+        )
+
+        return RolesList.model_validate(response)
```
mattgd, line 233: We don't use WorkOSListResource, because the API isn't paginated, but perhaps it'd be more standardized if it was?

jonatascastro12 (approved):

PR #398, opened 2025-01-13
Don't overwrite default transport so proxies can be set via environment variables

mattgd:
## Description
Don't overwrite default transport, so proxies can be set via environment variables. Tested and confirmed requests work as usual with no proxy, and are proxies properly when setting HTTPS_PROXY and ALL_PROXY environment variables.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/utils/http_client.py
+++ b/workos/utils/http_client.py
@@ -38,7 +38,7 @@ class SyncHTTPClient(BaseHTTPClient[httpx.Client]):
         client_id: str,
         version: str,
         timeout: Optional[int] = None,
-        transport: Optional[httpx.BaseTransport] = httpx.HTTPTransport(),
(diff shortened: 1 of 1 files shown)


workos/utils/http_client.py
```diff
@@ -51,7 +51,9 @@ def __init__(
             base_url=base_url,
             timeout=timeout,
             follow_redirects=True,
-            transport=transport,
+            # Only pass the transport if a custom one is provided, otherwise let httpx use
+            # the default so we don't overwrite environment configurations like proxies
+            transport=transport if transport else None,
```
mattgd, line 56: I can simplify this. Going to push up an update.

jonatascastro12 (approved):

PR #406, opened 2025-01-29
Add `intent_options` parameter on `Portal.generate_link()` method

jonatascastro12:
## Description

This PR introduces the `intent_options` parameter in the `Portal.generate_link()` method. This feature is not generally available on WorkOS API. Contact WorkOS support for access.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/workos/portal.py
+++ b/workos/portal.py
@@ -1,6 +1,7 @@
-from typing import Optional, Protocol
+from typing import Optional, Protocol, Dict, Literal, Union
 from workos.types.portal.portal_link import PortalLink
 from workos.types.portal.portal_link_intent import PortalLinkIntent
+from workos.types.portal.portal_link_intent_options import IntentOptions
 from workos.utils.http_client import SyncHTTPClient
 from workos.utils.request_helper import REQUEST_METHOD_POST
 
@@ -16,6 +17,7 @@ class PortalModule(Protocol):
         organization_id: str,
         return_url: Optional[str] = None,
         success_url: Optional[str] = None,
+        intent_options: Optional[IntentOptions] = None,
     ) -> PortalLink:
         """Generate a link to grant access to an organization's Admin Portal
 
@@ -47,13 +49,16 @@ class Portal(PortalModule):
         organization_id: str,
         return_url: Optional[str] = None,
         success_url: Optional[str] = None,
+        intent_options: Optional[IntentOptions] = None,
     ) -> PortalLink:
         json = {
             "intent": intent,
             "organization": organization_id,
             "return_url": return_url,
             "success_url": success_url,
+            "intent_options": intent_options,
         }
+
         response = self._http_client.request(
             PORTAL_GENERATE_PATH, method=REQUEST_METHOD_POST, json=json
         )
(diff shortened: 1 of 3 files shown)


workos/portal.py
```diff
@@ -16,6 +16,7 @@ def generate_link(
         organization_id: str,
         return_url: Optional[str] = None,
         success_url: Optional[str] = None,
+        intent_options_sso_bookmark_slug: Optional[str] = None,
```
mattgd, line 19: I wonder if we should defined this as a structured TypedDict so it makes future expansion easier.

For example:
```
# types/portal/portal_link_intent_options.py
class SSOIntentOptions(TypedDict):
    bookmark_slug: str


class IntentOptions(TypedDict):
    sso: SSOIntentOptions

# portal.py
def generate_link(
        self,
        *,
        intent: PortalLinkIntent,
        organization_id: str,
        return_url: Optional[str] = None,
        success_url: Optional[str] = None,
        intent_options: Optional[IntentOptions] = None,
    ) -> PortalLink:
    ...
    if intent_options and intent_options["sso"]:
        json["intent_options"] = {
            "sso": {
                "bookmark_slug": intent_options["sso"]["bookmark_slug"],
             }
         }
```

Usage would be something like:
```
Portal.generate_link(
    intent="sso",
    organization_id="org_123",
    intent_options={"sso": {"bookmark_slug": "test"}},
)
```

workos/portal.py
```diff
@@ -47,13 +48,22 @@ def generate_link(
...
     ) -> PortalLink:
-        json = {
+        json: Dict[str, Union[str, None, Dict[str, Dict[str, str]]]] = {
             "intent": intent,
             "organization": organization_id,
             "return_url": return_url,
             "success_url": success_url,
         }
+
+        if intent_options_sso_bookmark_slug:
```
mattgd, line 60: Probably not harmful, but should we only allow these options if `intent = sso`?

mattgd (commented):
Left a method signature suggestion. Let me know your thoughts.

workos/portal.py
mattgd: Ah interesting. Here's the autocomplete my editor gives for the example above with the TypedDict:
CleanShot 2025-01-30 at 09 09 32

mattgd: Down to proceed if you think it's alright, I just worry we could end up with a long list of parameters:
```
intent_options_sso_bookmark_slug
intent_options_sso_other_option
intent_options_dsync_role_mapping_option
etc.
```

jonatascastro12: I see. OK. I'll go forward with the `TypedDict`.

jonatascastro12: OK. I'm disappointed with PyCharm now.
CleanShot 2025-01-30 at 12 55 34

mattgd: Does PyCharm have a Pyright plugin?

mattgd (approved):
Looks good!

workos/portal.py
jonatascastro12: But never tried.

jonatascastro12: So is pyright the thing that does this?

mattgd: There may be other plugins, but yeah that's what's doing the type checking/auto complete in VSCode/Cursor.

PR #421, opened 2025-03-18
ron: adding support for context param

rarevalo13:
## Description
Adds support for the `context` param which is needed when you set the Login Endpoint URI
## Documentation
forthcoming documentation 

```
[ ] Yes
```

### DIFF
--- a/workos/user_management.py
+++ b/workos/user_management.py
@@ -345,6 +345,7 @@ class UserManagementModule(Protocol):
         code_challenge: Optional[str] = None,
         prompt: Optional[str] = None,
         screen_hint: Optional[ScreenHintType] = None,
+        context: Optional[str] = None,
     ) -> str:
         """Generate an OAuth 2.0 authorization URL.
 


workos/user_management.py
```diff
@@ -345,6 +345,7 @@ def get_authorization_url(
         code_challenge: Optional[str] = None,
         prompt: Optional[str] = None,
         screen_hint: Optional[ScreenHintType] = None,
+        context: Optional[str] = None,
```
mattgd, line 348: Confirming we don't intend to support this for standalone SSO as well?
https://github.com/workos/workos-python/blob/8def89f4fcd89414786f7a5fb32b1cf7667544e7/workos/sso.py#L49-L59

mattgd (approved):
Looks good, just one question.

workos/user_management.py
rarevalo13: I don't think so, `context` is related to the Login Endpoint under redirects, and only is set when you sent invites, or password resets, both things that are exclusive to UM. 
2025-Tuesday-19 05@2x

nholden:
Closing because we no longer require `context`!

PR #424, opened 2025-03-28
Don't arbitrarily constrain dependency versions

mrcljx:
## Description

For libraries, the best practice is to only list **known incompatibilities** and not _potential_ future incompatibilities. Using `<=` (or `~=`) constraints can:

- Lead to dependency resolution conflicts for downstream consumers
- Prevent users from using newer versions of dependencies that are likely still compatible
- Reduce composability in larger environments or monorepos

Note how `pydantic` itself has no [`<=` constraints](https://github.com/pydantic/pydantic/blob/main/pyproject.toml#L45). You'll find the same with other libraries.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-httpx~=0.28.1
-pydantic~=2.10.4
+httpx>=0.28.1
+pydantic>=2.10.4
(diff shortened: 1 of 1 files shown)


nicknisi (approved):
Thanks, @mrcljx!

PR #430, opened 2025-04-12
Add Support for a JWT Leeway Parameter

Th3R3p0:
## Description
This PR adds support for configuring a JWT leeway parameter that helps address token validation timing issues. When authenticating with WorkOS, occasional errors like "The token is not yet valid (iat)" can occur due to clock skew between servers. The leeway parameter creates a time buffer around token validation, allowing slightly expired or not-yet-valid tokens to be accepted.

The implementation adds a jwt_leeway parameter (default: 0) to the client configuration that's passed through to the JWT validation operations. Users can now specify a custom leeway value when creating a WorkOS client, and the setting will be consistently applied to all token operations. This change is fully backward compatible and includes tests to verify the leeway functionality works as expected with both slightly and significantly expired tokens.

## Documentation

```
[x] Yes
```

### DIFF
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -2,6 +2,7 @@ import pytest
 from unittest.mock import Mock, patch
 import jwt
 from datetime import datetime, timezone
+import time
 
 from tests.conftest import with_jwks_mock
 from workos.session import AsyncSession, Session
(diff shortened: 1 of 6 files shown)


nicknisi:
@Th3R3p0 IMO the code looks good and this looks like a worthwhile feature to add, thank you! It looks like CI is failing on formatting. Would you mind running the following and pushing up the fixes?

```bash
pipx run black .
```

Thanks!

PR #432, opened 2025-04-17
Allow organization_id to be optional on Connection for OAuth connections

mattgd:
## Description
Allow `organization_id` to be optional on `Connection` for OAuth connections.

## Documentation

```
[ ] Yes
```


gcarvelli (approved):

mattgd:
OAuth connection actually shouldn't be returned from the API in the first place, so investigating some more before I merge.

PR #434, opened 2025-04-17
Add custom_attributes field to SSO Profile

mattgd:
## Description
Add `custom_attributes` field to SSO Profile.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/tests/test_sso.py
+++ b/tests/test_sso.py
@@ -29,7 +29,7 @@ class SSOFixtures:
             idp_id="",
             first_name=None,
             last_name=None,
-            profile=None,
+            role=None,
             groups=None,
             raw_attributes={},
         ).dict()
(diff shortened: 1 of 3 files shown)


tests/test_sso.py
```diff
@@ -29,7 +29,7 @@ def mock_magic_link_profile(self):
             idp_id="",
             first_name=None,
             last_name=None,
-            profile=None,
+            role=None,
```
mattgd, line 32: This was incorrect in fixture, just fixing.

workos/types/sso/profile.py
```diff
@@ -23,7 +23,8 @@ class Profile(WorkOSModel):
     idp_id: str
     role: Optional[ProfileRole] = None
     groups: Optional[Sequence[str]] = None
-    raw_attributes: Mapping[str, Any]
+    custom_attributes: Optional[Mapping[str, Any]] = None
```
mattgd, line 26: I'd like to add a generic type for custom_attributes that developers can override, but ran into some mypy limitations with default values for generic types. Keeping this as-is for now.

gcarvelli (approved):

PR #440, opened 2025-05-01
Avoid decoding jwt twice

na-ka-na:
## Description
Currently the Session::authenticate() function (which runs on every request and consumes CPU cycles) is decoding the jwt twice unnecessarily. This small refactor fixes that

## Documentation

No


gcarvelli:
This change is released in release [v5.20.2](https://github.com/workos/workos-python/releases/tag/v5.20.2).

PR #491, opened 2025-11-05
Expose api key validation

denizs:
## Documentation

Once a new version is released, one could update the code examples that are currently limited to `authkit-nextjs` and `express`.

```
[x] Yes
```

### DIFF
--- a/tests/test_sso.py
+++ b/tests/test_sso.py
@@ -30,6 +30,7 @@ class SSOFixtures:
             first_name=None,
             last_name=None,
             role=None,
+            roles=None,
             groups=None,
             raw_attributes={},
         ).dict()
--- a/tests/test_user_management.py
+++ b/tests/test_user_management.py
@@ -1194,3 +1194,59 @@ class TestUserManagement(UserManagementFixtures):
         )
         assert request_kwargs["method"] == "post"
         assert isinstance(invitation, Invitation)
+
+    def test_resend_invitation(
+        self, capture_and_mock_http_client_request, mock_invitation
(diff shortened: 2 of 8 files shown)


denizs:
@mattgd - I'd like to get this contribution in, but I'm unsure of the next steps. The PR template tells me link a PR to corresponding docs, however, searching for `docs` in your org's repos didn't yield any result. Could please point me into the right direction?

workos/api_keys.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __init__(self, http_client: SyncHTTPClient):
+        self._http_client = http_client
+
+    def validate_api_key(self, *, value: str) -> ApiKey:
+        response = self._http_client.request(
+            API_KEY_VALIDATION_PATH, method=REQUEST_METHOD_POST, json={
+                "value": value}
+        )
+        return ApiKey.model_validate(response["api_key"])
```
nholden: When I passed an invalid value into this method, I saw this error message:

> Input should be a valid dictionary or instance of ApiKey

From [the docs](https://workos.com/docs/reference/authkit/api-keys/validate):

> If the key is invalid, the endpoint returns null for the api_key field.

Do we need to handle the invalid case differently?

nholden (approved):
Thank you, @denizs! This worked great for me. I'm going to see if I can loop in a Python expert for another review, but we should get this in a release relatively quickly.

denizs:
Thanks @nholden! And sorry for the slight delay - I was on vacation

PR #496, opened 2025-11-20
Move from setup.py to uv+pyproject.toml

birdcar:
## Description

I ran this by @nicknisi, but this is an upgrade to the plumbing behind our package building a testing to use UV instead of `setup.py`.

"Why?" I hear you asking; ultimately a few reasons.

- **Speed of development**: UV is fast as hell and can manage/install its own versions of python, which means testing against multiple versions of Python is easier / more straightforward (I will open some future PRs to make this a single command, in fact)
- **Keeping up with Python conventions**: There are still a lot of tutorials and old recommendations laying around that call for a `setup.py` file and direct invocation, but all direct invocations of `setup.py` have been essentially deprecated for years, and since PEP 518/621 landed, it's preferred to move away from the file entirely towards non-code files (namely, `pyproject.toml`).
- **Reproducibility**: Moving to a `src/` layout (i.e. the python package is in the `src` directory instead of the `root` directory) and managing its installation with `uv` also means we [avoid some common and easy to hit bugs](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/) when it comes to catching import problems etc.

There are other reasons I could probably elucidate, but I genuinely think this is a fairly uncontroversial change since a lot of the Python community is running to use UV as it stands.

Importantly, this does not change user workflows at all, with the minor exception of affecting those who were executing `python setup.py install` directly previously (I'm unclear if there's a way to know that up front. We can bump the version and communicate this as a breaking change if so).

## Documentation

```
[ ] Yes
```


birdcar:
Just realized I need to update the release workflow as well. Don't merge before I do that (should only take a moment or two)

birdcar:
The release workflow has been updated now, and [according to the UV docs](https://docs.astral.sh/uv/guides/package/#publishing-your-package) the way we should go about authentication for publishing a package is to [configure this release action as a trusted publisher](https://docs.pypi.org/trusted-publishers/adding-a-publisher/) on PyPI.

I can only make part of that process happen, namely, I can follow the guide for creating the release workflow, including the recommended environment (which I did in ffcc2a3). The rest will, I'm sure, need to be discussed or acted on by someone else (@nicknisi maybe?).

birdcar:
@gjtorikian I can absolutely split those out. I think my only concern there is that the *current* workflows will _for sure_ be broken after merge with this update. Having said that:

> IIRC, GitHub isn't running the ci.yml in this branch

Afaict GitHub is in fact running the updated CI workflows on this branch (note [the "UV" specific pieces in the completed checks](https://github.com/workos/workos-python/actions/runs/20925295639/job/60121416377?pr=496)).

The only workflow that should [need additional TLC is the release workflow](https://github.com/workos/workos-python/pull/496#issuecomment-3559078336).

birdcar:
> Just to move this along, how about splitting out the release.yml change, then working with $someone to a) make the PyPi change, b) make the GitHub environment change, and c) cut the release?

Can do. I'm out for the day so this will be a tomorrow priority ✌🏽

birdcar:
@gjtorikian I uno-reverse'd your uno-reverse. We can just use `uvx` to maintain the original package publishing step with `twine` while still moving everything else to UV, including the release build. Made the change in [b34f46c](https://github.com/workos/workos-python/pull/496/commits/b34f46cb2fcd4052add7ebf7fe03d02e2ebe7c09).

PR #517, opened 2026-01-15
ci: Migrate release workflow to Trusted Publishing

birdcar:
## Description

This PR migrates our PyPI release publishing from API token authentication to [Trusted Publishing (OIDC)](https://docs.pypi.org/trusted-publishers/adding-a-publisher/) for improved security. 

Additionally, while working on the PR I [reviewed the `uv` example workflow](https://github.com/astral-sh/trusted-publishing-examples/blob/main/.github/workflows/release.yml) and came to the realization that our previous release strategy merely duplicated the testing strategy from our CI tests (i.e. using `pytest` in a virtual environment) and **not** whether or not the package would be importable by clients and that all modules were accessible.

The release workflow now operates in isolated stages:

1. Build the package
2. Smoke test against each supported version of Python, failing the release if there's an issue
3. Upload the package to PyPI via Trusted Publishing token

## Documentation

```
[ ] Yes
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "workos"
-version = "5.39.0"
+version = "5.39.1"
 description = "WorkOS Python Client"
 readme = "README.md"
 license = "MIT"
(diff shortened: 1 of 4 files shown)


gjtorikian (approved):
thanks!

PR #524, opened 2026-02-01
Add custom_attributes field to OrganizationMembership

ajworkos:
## Summary

Adds `custom_attributes` field to `OrganizationMembership` model to expose custom attributes from identity providers.

## Changes

- Added `custom_attributes` field to OrganizationMembership type/model
- Field type: Record/Map/Dictionary of string keys to any values
- Always present, defaults to empty object `{}`
- Updated fixtures and tests

## API Field Details

The field will be present in:
- REST API responses
- Webhook events
- Events API responses

JSON field name: `custom_attributes` (snake_case)

### DIFF
--- a/src/workos/types/user_management/organization_membership.py
+++ b/src/workos/types/user_management/organization_membership.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Literal, Sequence, Optional
+from typing import Any, Literal, Mapping, Optional, Sequence
 from typing_extensions import TypedDict
 
 from workos.types.workos_model import WorkOSModel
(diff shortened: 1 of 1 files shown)


src/workos/types/user_management/organization_membership.py
```diff
@@ -21,5 +21,6 @@ class OrganizationMembership(WorkOSModel):
     role: OrganizationMembershipRole
     roles: Optional[Sequence[OrganizationMembershipRole]] = None
     status: LiteralOrUntyped[OrganizationMembershipStatus]
+    custom_attributes: Dict[str, Any]
```
mattgd: We've been preferring to use `Mapping` instead of `Dict` for these read-only cases:
```suggestion
    custom_attributes: Mapping[str, Any]
```

PR #556, opened 2026-02-19
feat: fga adding base types and module registraiton

swaroopAkkineniWorkos:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,7 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "Bash(gh api:*)"
-    ]
-  }
-}


src/workos/types/authorization/access_evaluation.py
```diff
@@ -0,0 +1,7 @@
+from workos.types.workos_model import WorkOSModel
+
+
+class AccessEvaluation(WorkOSModel):
```
atainter: Why not just call this a CheckResponse?

src/workos/types/user_management/organization_membership_status.py
```diff
@@ -0,0 +1,3 @@
+from typing import Literal
+
+OrganizationMembershipStatus = Literal["active", "inactive", "pending"]
```
atainter: It's risky to duplicate this if it already exists in another package. Could drift from the other types.

src/workos/types/authorization/organization_membership.py
```diff
@@ -0,0 +1,27 @@
...
+    authorization endpoints may omit it.
+    """
+
+    object: Literal["organization_membership"]
+    id: str
+    user_id: str
+    organization_id: str
+    organization_name: str
+    status: LiteralOrUntyped[OrganizationMembershipStatus]
+    custom_attributes: Optional[Mapping[str, Any]] = None
```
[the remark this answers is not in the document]
swaroopAkkineniWorkos: i included it because it's on the `UserlandUserOrganizationMembershipBase` object that gets returned here  https://github.com/workos/workos/blob/61dd6210464228608b3f0013993ed6279051fa73/packages/api/src/authorization-resources/authorization-resources.controller.ts#L432

src/workos/types/user_management/organization_membership_status.py
swaroopAkkineniWorkos: I actually broke it out out [organization_membership.py](https://github.com/workos/workos-python/pull/556/changes#diff-b58be39a61de205c1d803da04ff0a9fd46bf755860d6b1afc4308eb374463aab) into it's own file so it could be imported as the single source of truth

atainter (approved):
Looks good to me. Should probably get a review from python code owners as well

PR #571, opened 2026-02-20
feat: FGA_5 list_resources_for_membership, list_memberships_for_resource, list_memberships_for_resource_by_external_id

swaroopAkkineniWorkos:
…list_memberships_for_resource_by_external_id

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/workos/authorization.py
+++ b/src/workos/authorization.py
@@ -3,7 +3,6 @@ from functools import partial
 from typing import Any, Dict, Optional, Protocol, Sequence, Union
 
 from pydantic import TypeAdapter
-from typing_extensions import TypedDict
 
 from workos.types.authorization.access_check_response import AccessCheckResponse
 from workos.types.authorization.assignment import Assignment


src/workos/authorization.py
```diff
@@ -712,3 +904,127 @@ async def add_environment_role_permission(
...
+        organization_id: str,
+        resource_type_slug: str,
+        external_id: str,
+        *,
+        permission_slug: str,
+        assignment: Optional[Literal["direct", "indirect"]] = None,
+        limit: int = DEFAULT_LIST_RESPONSE_LIMIT,
+        before: Optional[str] = None,
+        after: Optional[str] = None,
+        order: PaginationOrder = "desc",
```
swaroopAkkineniWorkos: Would ideally like to move this to a PAginationOptions object, but want to keep things consistence with above

src/workos/authorization.py
```diff
@@ -58,16 +65,25 @@ class ResourceListFilters(ListArgs, total=False):
...
+    AuthorizationResource, ResourcesForMembershipListFilters, ListMetadata
+]
 
 
-class ParentResourceByExternalId(TypedDict):
-    parent_resource_external_id: str
-    parent_resource_type_slug: str
+class AuthorizationOrganizationMembershipListFilters(ListArgs, total=False):
+    permission_slug: str
+    assignment: Optional[Literal["direct", "indirect"]]
```
atainter: Might be worth making `assignment` a type since we'll most likely add it to the list resources for membership endpoint soon.

PR #607, opened 2026-04-10
feat: Add resource_type_slug to createOrganizationRole to create resource-scoped custom roles

atainter:
Also makes permission slug optional. If omitted, the service will create auto-generated slugs


swaroopAkkineniWorkos (approved):

gjtorikian:
:wave: This needs to be regenerated against the new OpenAPI spec change. I can handle that and ping you for review.

PR #610, opened 2026-04-13
fix: regenerate methods to include missing params

gjtorikian:
## Description

v6 wasn't regenerated with the latest and greatest spec; this PR adds some newly introduced params.

### DIFF
--- a/src/workos/user_management/_resource.py
+++ b/src/workos/user_management/_resource.py
@@ -157,6 +157,9 @@ class UserManagement:
         email: str,
         password: str,
         invitation_token: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


src/workos/user_management/_resource.py
```diff
@@ -2285,19 +2356,27 @@ async def authenticate_with_password(
     async def authenticate_with_code(
```
dhimmel, line 2356: Should `authenticate_with_code_pkce` also get `ip_address`, `device_id`, `user_agent`?

dhimmel: opened https://github.com/workos/workos-python/issues/618

PR #625, opened 2026-04-16
fix: add documentation on sealing sessions

gjtorikian:
This PR adds documentation/guidance around applications that previously passed `session=` into `authenticate_with_code()` or `authenticate_with_refresh_token()`.

`user` was also erroneously set to `Optional` in these helper methods.


tiagoskaneta:
for reference this is still wrong in https://workos.com/docs/authkit/vanilla/python#save-the-encrypted-session, took me a while to figure out this given that I had no idea I had to check a migration guide to install the first time...

gjtorikian:
Ah dang -- will take a look there soon. Thanks for the feedback!

PR #633, opened 2026-04-24
fix: list Slack as an auth provider for the authorization URL

tgwizard:
## Description

Slack wasn't in here, but it works. I want to use it in `workos_client.user_management.get_authorization_url`. (This list seems somewhat inconsistent with other lists).

## Documentation

```
[x] No
```


gjtorikian:
Hey, thanks for this!

These files are now automatically generated by our OpenAPI spec and completely machine generated (`This file is auto-generated by oagen.`) -- so I updated that YAML and produced the full complete list here. This'll be out in the next patch.

PR #671, opened 2026-06-17
feat: Add roles field to OrganizationMembership models

ethinallen:
## Description

The `OrganizationMembership` model only exposed `role: SlimRole` (singular), but the API also returns `roles: List[SlimRole]` for multiple-role support. The event model (`OrganizationMembershipCreatedData`) already had this field — this PR adds it to the two remaining models:

- `OrganizationMembership` — used by `list_organization_memberships`, `get_organization_membership`, `create_organization_membership`, etc.
- `UserOrganizationMembershipBaseListData` — used by authorization and groups listing endpoints

The field is `Optional[List[SlimRole]]` (defaults to `None`), matching the existing pattern in `OrganizationMembershipCreatedData`. Backward compatible — `role` is unchanged, and `roles` gracefully handles absence from older API responses.

## Documentation

```
[ ] Yes
```

Link to Devin session: https://app.devin.ai/sessions/896e93121dcd42839221cfd4cebfec33
Requested by: @ethinallen

### DIFF
--- a/src/workos/common/models/user_organization_membership_base_list_data.py
+++ b/src/workos/common/models/user_organization_membership_base_list_data.py
@@ -6,10 +6,11 @@ from dataclasses import dataclass
 from datetime import datetime
 from enum import Enum
 from typing import cast
-from typing import Any, Dict, Literal, Optional
+from typing import Any, Dict, List, Literal, Optional
 from workos._types import _raise_deserialize_error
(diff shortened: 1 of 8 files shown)


gjtorikian:
Since these files are autogenerated, we need to determine if there's a problem in the OpenAPI spec representation, or the SDK generation. Otherwise, these edits will be deleted the next time the SDK is autogenerated.

PR #716, opened 2026-08-25
feat(generated): Add flag target resources

Deborah-Digges:
## Summary
- expose generated clients for flag targets and custom flag target types
- add the generated request and response models for both resources
- add focused fixtures and coverage for CRUD, pagination, and model round trips
- keep the generated manifest and top-level client synchronized with AUTH-6891

Requested by deborah.digges@workos.com via TARS


Deborah-Digges:
Closing this because there is a different paved path to generate SDK PRs

PR #725, opened 2026-09-05
feat: Add optional jwt_issuer option for session token validation

m0tzy:
## Summary

Adds an opt-in `iss` claim check to `Session` / `AsyncSession` access-token validation (both the `authenticate()` and `refresh()` `jwt.decode` calls). When `jwt_issuer` is configured on the client it is passed to PyJWT; when unset, behavior is unchanged (no `iss` check — PyJWT skips `_validate_iss` when `issuer is None`).

```python
WorkOSClient(..., jwt_issuer: Optional[Union[str, Sequence[str]]] = None)
AsyncWorkOSClient(..., jwt_issuer=...)
# falls back to WORKOS_ISSUER, comma-separated for a list

jwt.decode(access_token, key, algorithms=..., options={"verify_aud": False},
           leeway=self._client._jwt_leeway,
           issuer=self._client._jwt_issuer)   # None -> not validated
```

`jwt_issuer` accepts a single issuer or a list so an app can accept tokens from more than one issuer (e.g. during an issuer migration); PyJWT's `issuer` parameter already accepts `str | Container[str]`. A `WORKOS_ISSUER` env value is split on commas (trimmed, empties dropped; a single value stays a `str`). Named `jwt_issuer` to sit alongside the existing `jwt_leeway` option.

Only hand-maintained (`@oagen-ignore-file`) files are touched: `_base_client.py`, `session.py`.

Opt-in rather than defaulted because the API does not mint a single issuer shape — it varies by environment (`https://api.workos.com` for legacy environments, `https://api.workos.com/user_management/<clientId>` for environments created since mid-2025, custom auth domains, and flag-gated path variants). Mirrors workos/workos-node#1694.

Link to Devin session: https://app.devin.ai/sessions/0ee38e859a9849658a7cdb2d215d89a6
Open in Devin Desktop: https://app.devin.ai/desktop/session/0ee38e859a9849658a7cdb2d215d89a6?variant=devin
Requested by: @m0tzy


gjtorikian (approved):
Thanks!

PR #731, opened 2026-09-16
feat: switch default HTTP client to httpx2 and accept a custom http_client

### LINKED ISSUE #730: Replace httpx dependency with httpx2, add optional aiohttp support
### Summary
Currently the WorkOS Python SDK depends directly on `httpx` for making HTTP requests. This issue proposes switching the default HTTP client to `httpx2`, while also making it possible to optionally use `aiohttp` as an alternative backend.

httpx has become effectively unmaintained — no releases since 2024, issues and discussions are being closed without resolution. The ecosystem is moving to httpx2, the successor maintained by Pydantic Services Inc. with the original author (Tom Christie). Starlette already migrated (https://github.com/encode/starlette/pull/3291, merged May 2026), with the PR author noting:

    "httpx has become somehow unmaintained (closing issues, discussions, and no release since 2024), and [httpx2] is the least annoying path forward for every consumer of that package."

### Motivation
- Avoid being tightly coupled to a single HTTP library (`httpx`), which can cause dependency conflicts for consumers who use a different HTTP stack.
- `httpx2` is being proposed as the new default client library.
- Some users may prefer or already depend on `aiohttp` for async workloads, so offering it as an optional backend would reduce friction for those users.

### Proposed changes
1. Remove the hard dependency on `httpx` and replace it with `httpx2` as the default HTTP client used internally.
2. Introduce an HTTP client abstraction/interface so the underlying library (`httpx2` or `aiohttp`) can be swapped without touching business logic.
3. Add `aiohttp` as an optional extra (e.g. `pip install workos[aiohttp]`) for users who want to use it instead of `httpx2`.
4. Update sync/async request handling to work correctly with whichever backend is selected.
5. Update documentation and README to describe how to select an HTTP backend.

gjtorikian:
## Summary

Replaces the `httpx` dependency with `httpx2` (the Pydantic-stewarded fork of httpx 0.28.1 with an identical API) and puts the HTTP layer behind a small protocol so callers can pass their own client.

- `httpx2~=2.13` replaces `httpx~=0.28` in `dependencies`. `httpx` is no longer installed transitively.
- New hand-maintained `src/workos/_http.py`: `HTTPResponse`, the `HTTPBackend` / `AsyncHTTPBackend` protocols, `TransportError` / `TransportTimeout` / `TransportConnectError`, and one adapter that serves both `httpx2` and `httpx` 0.28 clients. All exported from `workos`.
- `WorkOSClient`, `AsyncWorkOSClient`, and `create_public_client` take `http_client=`. Accepts an httpx2 client, an httpx client, or any object implementing the protocol. A wrong-flavor client (sync vs async) raises a `TypeError` that names the fix. The SDK closes only clients it created.
- Query and JSON body encoding move into `_base_client.py` so every backend sends identical bytes (verified byte-for-byte against httpx2).
- Tests: `pytest-httpx` pins `httpx==0.28.*` and cannot intercept httpx2. It is replaced by a hand-maintained `httpx_mock` fixture in `tests/conftest.py` that preserves the API the 24 oagen-generated test files use. Those files are untouched.
- README gains an "HTTP Backends" section, including a reference aiohttp adapter labelled as an example.

Closes #730. The aiohttp extra requested there is intentionally not shipped: `httpx2` stays a hard dependency for the sync client and the default, so an in-tree aiohttp adapter would only buy session reuse at the cost of an optional-dependency matrix. The protocol makes it a small add-on if demand appears.

## Why this is a minor release

`httpx` never appeared in a public signature: the constructors took no HTTP client, and network errors were already rethrown as `WorkOSTimeoutError` / `WorkOSConnectionError` / `WorkOSError`. `docs/V6_MIGRATION_GUIDE.md` states raw httpx exceptions are not part of the contract. Dependency changes have shipped as non-breaking here before (cryptography v48, pyjwt 2.12).

## Compatibility notes

For code that only calls the SDK this is transparent: same requests, retries, and exceptions.

One runtime change: httpx2 verifies TLS against the operating system trust store (`truststore`) instead of certifi's bundle. `SSL_CERT_FILE` and `SSL_CERT_DIR` are honored exactly as before. Only an image with no system CA bundle at all is affected.

| If you... | Then... | Fix |
| --- | --- | --- |
| mock WorkOS calls in tests with `pytest-httpx` or `respx` | mocks no longer intercept | `WorkOSClient(..., http_client=httpx.Client())` in test setup |
| `import httpx` without declaring it | missing on fresh installs or a re-lock (an in-place `pip install -U workos` keeps it) | declare `httpx` |
| pin `anyio<4.10` or `idna<3.18` | resolver error at install time | relax the pin |
| install from a curated mirror | `httpx2`, `httpcore2`, `truststore` need approval | approve them, or stay on 10.3 |
| reach into `client._client` or filter the `httpx` logger | `AttributeError`; request logs now come from logger `httpx2` | use `http_client=`; filter `httpx2` |

## Test plan

- [x] `ruff format --check` and `ruff check`: clean
- [x] `pyright` (src and tests, strict): 0 errors
- [x] `pytest`: 2820 passed, generated test files unchanged
- [x] `uv build` then `tests/smoke_test.py` against the wheel in an isolated env: 11/11, `httpx` not installed
- [x] New `tests/test_http_backends.py`: protocol fake (retry, 4xx never retried, error mapping, close ownership), httpx adapter parametrized over `httpx2` and `httpx`, resolution errors, encoding parity against httpx2


src/workos/_http.py
```diff
@@ -0,0 +1,315 @@
...
+        method: str,
+        url: str,
+        *,
+        headers: Dict[str, str],
+        content: Optional[bytes],
+        timeout: float,
+    ) -> HTTPResponse:
+        try:
+            response = self._client.request(
+                method, url, headers=headers, content=content, timeout=timeout
```
birdcar: This current construction would mean that the SDK's query would be overridden entirely when a user-supplied client has default `params`, right? Both `httpx` and `httpx2` replace the query embedded in `url` with the client's defaults here, rather than merging them. For example:

```python
transport = httpx2.Client(params={"limit": 100})
client = WorkOSClient(api_key="sk_...", http_client=transport)
client.user_management.list_users(email="alice@example.com")
```

From my brief look, this would only send `?limit=100` and would silently drop the email filter (and the SDK's default order). I think that pagination cursors would be dropped the same way, so auto-pagination would keep fetching the first page.

It's possible we don't care and this is a "if you're gonna use a custom transport then you should explicitly pass every param" thing.

src/workos/_http.py
```diff
@@ -0,0 +1,315 @@
...
+        content: Optional[bytes],
+        timeout: float,
+    ) -> HTTPResponse: ...
+
+    async def close(self) -> None: ...
+
+
+if TYPE_CHECKING:
+    SyncHTTPClient = Union[httpx2.Client, httpx.Client, HTTPBackend]
+    AsyncHTTPClient = Union[httpx2.AsyncClient, httpx.AsyncClient, AsyncHTTPBackend]
```
birdcar: Super small nit: could these aliases also be defined at runtime, and imported normally in `_base_client.py`? They _currently_ appear in public annotations, but both their definitions and imports are guarded by `TYPE_CHECKING`. As a result:

```python
from typing import get_type_hints
from workos import WorkOSClient

get_type_hints(WorkOSClient.__init__)
# NameError: name 'SyncHTTPClient' is not defined
```

The async constructor similarly fails on `AsyncHTTPClient` (i.e.  both constructors' annotations resolve before this change). Normal SDK calls still work, but runtime annotation consumers would error if I'm thinking about this correctly.

Again, maybe not a concern we want to block shipping this change for, but something worth considering.

birdcar (commented):
In general I say "ship it", there are two nits that I don't think are blocking so I'm dropping this as a comment and can flip to approval whenever you say the word.

